PR #13, opened 2018-05-23
Correct header logo dimensions

ruivieira:
The logo looks squashed with the current 30% width / 30% height dimensions.
Removed the width so that the logo keeps its original aspect ratio.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -6,7 +6,7 @@
 </div>
 <div align="center">
     <br />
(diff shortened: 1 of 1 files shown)


tombentley:
@ruivieira thanks spotting this and for the PR. It still didn't render properly in Firefox (at least on my machine), so in this instance I have made and committed some CSS changes to fix it. Thanks again.

PR #32, opened 2018-08-31
Move Strimzi Roadmap to GitHub project

scholzj:
As discussed earlier ... I think the _Roadmap_ page is not well maintained and not easy to read. I tried to create a [GitHub project](https://github.com/strimzi/strimzi-kafka-operator/projects/1) to track the roadmap and the "epics and link to it from strimzi.io. Please let me know what you think.


tombentley:
:+1: on removing the existing roadmap and linking out to something else.

My main reservation is just that the Github Project looks strange:

* Do we really need to show columns for even very old releases (the columns are wider than my monitor)?
* Wouldn't it be more natural to have the columns in the reverse order (time moves to the right in my mind)?

scholzj:
So where would you put the _To Do_ column? Logically the _To Do_ column should be next to the newest / upcoming release. So would _To Do_ be the most right column?

ppatierno:
@scholzj yes but ... I don't think that it's possible to have the browser opening the roadmap and moving the scroll bar to the right (showing the most recent releases and "to do") so I think that the current layout is better. When the roadmap is opened, the people can see immediately what we are working on (release in process) and the to-do.

scholzj:
I merged this because the roadmap on the website was getting outdated. Any further changes to how the roadmap project looks and what are the columns could be addresses separately.

PR #45, opened 2018-11-27
Update Vault blog to include information about the CA chain

chris-vest:
@ppatierno Added corrections according to latest learnings. 🍂

### DIFF
--- a/_posts/2018-11-16-using-vault-with-strimzi.md
+++ b/_posts/2018-11-16-using-vault-with-strimzi.md
@@ -96,10 +96,10 @@ $ vault secrets enable pki
 #   of 87600 hours (10 years):
 $ vault secrets tune -max-lease-ttl=87600h pki
 
-# Generate the root CA, extracting the root CA's certificate to CA_cert.crt; the secret
+# Generate the root CA, extracting the root CA's certificate to root.crt; the secret
 #   key is not exported!
 $ vault write -field=certificate pki/root/generate/internal common_name="example.com" \
-       ttl=87600h > CA_cert.crt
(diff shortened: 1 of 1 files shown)


_posts/2018-11-16-using-vault-with-strimzi.md
```diff
@@ -166,7 +171,7 @@ kubectl create secret -n ${strimziNamespace} generic ${clusterName}-cluster-ca \
 
 # Certificate
 kubectl create secret -n ${strimziNamespace} generic ${clusterName}-cluster-ca-cert \
-  --from-file=ca.crt=intermediate.cert.pem \
+  --from-file=ca.crt=intermediate.chain.pem \
```
ppatierno: what is done here let me think that you are providing a certs chain to Strimzi as Cluster CA cert which sounds strange, at least to me :-)
Now Strimzi is going to sign the autogenerated ZK and K brokers/nodes certificates using this PEM file containing a chain. To be honest I have never done something like that. Isn't it enough to generate the Secret passing the intermediate CA cert and the Strimzi will sign brokers/nodes certs with that one? @tombentley @scholzj wdyt?

scholzj: If it works? Keep in mind that this is just the public key with the chain. I assume there is no chain for the private key which should IMHO make it ok.

ppatierno: Yeah I said a stupid one. Certs will be signed by related intermediate CA private key not by public certs in the chain.

PR #52, opened 2019-03-19
Add message producer and consumer to the Quickstarts

scholzj:
The Quickstarts are not making it clear how to send or receive messages. I added a new chapter about this. It is basically a copy of the corresponding section from the docs.

### DIFF
--- a/quickstarts/minikube/index.md
+++ b/quickstarts/minikube/index.md
@@ -56,4 +56,18 @@ my-cluster-zookeeper-2                        2/2     Running   0          2m30s
 strimzi-cluster-operator-78f8bf857-kpmhb      1/1     Running   0          3m10s
 ```
 
+# Send and receive messages
+
+Once the cluster is running, you can run a simple producer to send messages to Kafka topic (the topic will be automatically created):
+
+```shell
+kubectl run kafka-producer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list my-cluster-kafka-bootstrap:9092 --topic my-topic
+```
+
+And to receive them:
+
+```shell
+kubectl run kafka-consumer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --topic my-topic --from-beginning
+```
+
 Enjoy your Apache Kafka cluster, running on Minikube!
--- a/quickstarts/okd/index.md
+++ b/quickstarts/okd/index.md
@@ -53,4 +53,18 @@ my-cluster-zookeeper-2                        2/2     Running   0          2m30s
 strimzi-cluster-operator-78f8bf857-kpmhb      1/1     Running   0          3m10s
 ```
 
+# Send and receive messages
+
+Once the cluster is running, you can run a simple producer to send messages to Kafka topic (the topic will be automatically created):
+
+```shell
+oc run kafka-producer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list my-cluster-kafka-bootstrap:9092 --topic my-topic
+```
+
+And to receive them:
+
+```shell
+oc run kafka-consumer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --topic my-topic --from-beginning
+```
+
 Enjoy your Apache Kafka cluster, running on OKD!


scholzj:
@ppatierno I guess I leave that part for some other PR ;-)

scholzj:
@tombentley This does not piggy-back on the broker pods. It starts its own pods where the consumer / producer are run. Of course this is not a way how you would run your real application in production, but I'm not sure this is the right place to teach them about Kubernetes basics. IMHO the main point of this is to show them that it works and how to connect (i.e. which service / port).

tombentley:
Sure, I'm not saying we should pontificate about it, just a sentence that this is a quick and simple way to check that it works.

quickstarts/minikube/index.md
```diff
@@ -56,4 +56,18 @@ my-cluster-zookeeper-2                        2/2     Running   0          2m30s
...
+```shell
+kubectl run kafka-producer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list my-cluster-kafka-bootstrap:9092 --topic my-topic
+```
+
+And to receive them:
+
+```shell
+kubectl run kafka-consumer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --topic my-topic --from-beginning
+```
+
```
tombentley:
```suggestion

These commands are a quick and simple way to check that the brokers are running correctly, without needing to use a `Deployment` as you would typically use for real producers and consumers.
```

quickstarts/okd/index.md
```diff
@@ -53,4 +53,18 @@ my-cluster-zookeeper-2                        2/2     Running   0          2m30s
...
+```shell
+oc run kafka-producer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list my-cluster-kafka-bootstrap:9092 --topic my-topic
+```
+
+And to receive them:
+
+```shell
+oc run kafka-consumer -ti --image=strimzi/kafka:0.11.1-kafka-2.1.0 --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --topic my-topic --from-beginning
+```
+
```
tombentley:
```suggestion

These commands are a quick and simple way to check that the brokers are running correctly, without needing to use a `Deployment` as you would typically use for real producers and consumers.
```

scholzj: So what is our answer for the obvious question _How to use a `Deployment` as you would typically use for real producers and consumers?_ 

My feeling is that if we say this, we should link somewhere for the real thing. Without that it feels a lot like _We wanted to give you here the real thing but we don't have it. So instead we give you just this cheap hack._.

tombentley: Well we ought to have that documentation, don't you think? And if we don't have it we should write it. But that's really outside the scope of this PR.

scholzj: tBH, my assumption would be that most people who come to the Strimzi Quickstart Guide will:
* Have a basic knowledge of Kafka
* Have a basic knowledge of Kubernetes
Because why would you otherwise go to a project about running Kafka on Kubernetes? And in that case this should be all obvious. They will know that they should use deployment for their actual application and that it will not use the console consumer and producer.

You don't seem to believe in that. So which part do you think is not obvious and needs explaining? Is it the Kubernetes part or the Kafka part? Or both? For the Kubernetes, if you think it's not obvious, we should add link to some Kubernetes basics docs explaining deployments. For the Kafka part we should add a link about using the Kafka Consumer and Producer APIs or something.

That will tell the users where to continue if they didn't know that already and don't leave this as some unfinished point.

tombentley: I thought we had a plan to create some documentation for how to write "your first Kafka producer" and "your first Kafka consumer":

* How to find out the listener address to use configuring the client
* Briefly explaining using the Java API
* Explaining how to run that as a Deployment.

I suppose there are really two cases of each (within and outside the cluster), but that's where I would image any such link pointing to.

I suppose I don't have a problem with not having this extra sentence in this PR while we lack this documentation.

scholzj: TBH, the first one should be IMHO covered already today. The other two IMHO do not belong to our docs. One should be in Kubernetes / OpenShift docs and the other in Kafka docs. AFAIK the Kafka clients docs we talked about were mainly a downstream thing. But I think that if you think such docs are needed, we don't have to wait for them. What about linking here [https://kubernetes.io/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro/](https://kubernetes.io/docs/tutorials/kubernetes-basics/deploy-app/deploy-intro/) and here [http://kafka.apache.org/documentation/#api](http://kafka.apache.org/documentation/#api)? Because I'm a bit sceptical about when will we get to writing such docs.

We can also just merge this and leave that for other PRs.

ppatierno: I think that we should mention that this is a way for testing that the cluster is running but not mentioning about `Deployment` and Kubernetes stuff.
Developers could be interested to access the running cluster from outside so running the Kafka client on their laptop and not on the Kubernetes cluster.

PR #66, opened 2019-06-07
Move old docs to archive and add section for bridge

scholzj:
The bridge API docs are now built on master and pushed to the website. So we should link them in the _Documentation_ page. To make that more visible, I also created a _Documentation Archive_ page and move all the older docs there.


scholzj:
@tombentley strimzi/strimzi-kafka-operator#1697

PR #84, opened 2019-08-22
Wrong namespace in the commands to run receiver and producer

dellaglio:
The producer and receiver commands use the kafka namespace, while the kafka cluster in created in myproject namespace. I edited the producer and consumer commands replacing the kafka namespace with myproject.


scholzj:
Thanks a lot for noticing the issue and taking the time for the PR @dellaglio

PR #88, opened 2019-09-04
Added CNCF logo in the homepage

ppatierno:


sjwoodman:
Generally LGTM. I wonder if it would be better higher up the page?

However, given that we are working on a redesigned site I think adding the logo in this way is ok.

PR #90, opened 2019-09-09
Update links to mailing lists and Slack

scholzj:
This PR upúdates the mailing list and Slack links to the new CNCF hosted links.

### DIFF
--- a/contributing/index.md
+++ b/contributing/index.md
@@ -20,5 +20,5 @@ The [Documentation Contributor Guide](https://strimzi.io/contributing/guide/) de
 
 If you want to get in touch with us first before contributing, you can use:
 
-* [Strimzi mailing list](https://www.redhat.com/mailman/listinfo/strimzi)
-* [Strimzi Slack workspace](https://join.slack.com/t/strimzi/shared_invite/enQtMzU2Mjk3NTgxMzE5LTYyMTUwMGNlMDQwMzBhOGI4YmY4MjhiMDgyNjA5OTk2MTFiYjc4M2Q3NGU1YTFjOWRiMzM2NGMwNDUwMjBlNDY)
+* [#strimzi channel on CNCF Slack](https://slack.cncf.io/)
+* [Strimzi Dev mailing list](https://lists.cncf.io/g/cncf-strimzi-dev/topics)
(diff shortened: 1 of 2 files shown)


contributing/index.md
```diff
@@ -20,5 +20,5 @@ The [Documentation Contributor Guide](https://strimzi.io/contributing/guide/) de
 
 If you want to get in touch with us first before contributing, you can use:
 
-* [Strimzi mailing list](https://www.redhat.com/mailman/listinfo/strimzi)
-* [Strimzi Slack workspace](https://join.slack.com/t/strimzi/shared_invite/enQtMzU2Mjk3NTgxMzE5LTYyMTUwMGNlMDQwMzBhOGI4YmY4MjhiMDgyNjA5OTk2MTFiYjc4M2Q3NGU1YTFjOWRiMzM2NGMwNDUwMjBlNDY)
+* [#strimzi channel on CNCF Slack](https://slack.cncf.io/)
```
ppatierno: This is the link that CNCF seems to use in their sendbox projects page:

https://cloud-native.slack.com/messages/strimzi

index.md
```diff
@@ -42,8 +42,8 @@ Documentation for all releases can be found in the [Documentation](/documentatio
 
 If you encounter any issues while using Strimzi, you can get help using:
 
-* [Strimzi mailing list](https://www.redhat.com/mailman/listinfo/strimzi)
-* [Strimzi Slack workspace](https://join.slack.com/t/strimzi/shared_invite/enQtMzU2Mjk3NTgxMzE5LTYyMTUwMGNlMDQwMzBhOGI4YmY4MjhiMDgyNjA5OTk2MTFiYjc4M2Q3NGU1YTFjOWRiMzM2NGMwNDUwMjBlNDY)
+* [#strimzi channel on CNCF Slack](https://slack.cncf.io/)
```
ppatierno: ditto as above.

contributing/index.md
scholzj: That links is useless unless you are already registered. So I think we should keep what I have there.

index.md
scholzj: That links is useless unless you are already registered. So I think we should keep what I have there.

PR #96, opened 2019-09-30
Add 0.14.0 release docs

scholzj:
This PR updates the website for the 0.14.0 release.


charts/index.yaml
```diff
@@ -1,6 +1,31 @@
 apiVersion: v1
 entries:
   strimzi-kafka-operator:
+  - apiVersion: v1
+    appVersion: 0.14.0
+    created: 2019-09-29T08:51:10.73893-07:00
+    description: 'Strimzi: Kafka as a Service'
```
ppatierno: Don't we want to use "Strimzi: Apache Kafka running on Kubernetes"?

scholzj: I guess we can change it if you want, but please do it in a separate PR. This PR really just adds the new version and nothing else.

ppatierno: ok

PR #112, opened 2019-11-12
Use absolute address to the log

scholzj:


_includes/header.html
```diff
@@ -10,7 +10,7 @@
 </script>
 <div align="center">
     <br />
-    <img alt="Strimzi" src="assets/strimzi.png" />
+    <img alt="Strimzi" src="https://strimzi.io/assets/strimzi.png" />
```
ppatierno: isn't it enough just `/assets/strimzi.png`?

scholzj: Might ... but maybe that will bubble out somwehere else. This sjhould be bulletproof :-)

ppatierno: Why not trying then? we have this cool preview feature thanks to netlify :-)

scholzj: I should have left you do it. We had the preview before as well ... and how much did it helped?

ppatierno: I forgot about it before :P

PR #119, opened 2020-01-17
Fix documentation archive to support multiple books

scholzj:
The documentation archive doesnt currently contain links to the new Overview and Quick Start books added in new releases. This Pr tries to fix that.


scholzj:
This is the updated page: https://deploy-preview-119--strimzi.netlify.com/documentation/archive/

PR #125, opened 2020-03-09
Using MirrorMaker 2.0

PaulRMellor:

Post to describe using MirrorMaker 2.0


PaulRMellor:
First draft.
Concepts and config.
Going to ask Andrew Borley to take a look, with a view to adding some of the technical details we might need.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+MirrorMaker 2.0 **_connectors_** -- remember, we're based on Kafka Connect now -- and related **_internal topics_** help manage the transfer and synchronization of data between the clusters.
+
+Different to the previous version of MirrorMaker, radically different, but different doesn't mean more complicated here.
+In fact, once you know the essentials, setting up is rather straightforward.
+
+Using Strimzi, you just need to work out how you want to configure a `KafkaMirrorMaker2` resource, which will define your MirrorMaker 2.0 deployment.
+We'll look a bit more at configuring the `KafkaMirrorMaker2` resource later on.
+
+First, something you'll be keen to know.
+Is the sequal better than the original?
```
dalelane: minor typo: sequal/sequel

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+The target cluster had to be manually maintained to match topic configuration changes in the source cluster.
+
+This time the topic configuration is now automatically synchronized between source and target clusters through configuration properties.
+Configuration changes are propagated to remote topics so that new topics and partitions are detected and created.
+The need for rebalancing vanishes.
+
+### Offset tracking and mapping
+
+In the old version of MirrorMaker, the offset of the source topic in the target cluster begins when the replication begins.
+The `__consumer-offset topic` is not mirrored.
```
dalelane: The topic name here should be `__consumer_offsets` and I think that perhaps “topic” was meant to be outside the code markup?

PaulRMellor: Yes. You're right. Updated.

PaulRMellor:
> hi @PaulRMellor - I’m a colleague of @ajborley’s
> 
> Just a quick comment to share a couple of tiny things I noticed, hope that’s okay. Looks great otherwise!

Thanks for checking Dale. If you think we need to say anything more, please let me know.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+Strategies such as using timestamps can be adopted, but it adds complexity.
+
+Poof! These issues entirely disappear with MirrorMaker 2.0.
+Instead, we get simplicity and sophistication through the `MirrorCheckpointConnector`.
+
+`MirrorCheckpointConnector` tracks and maps offsets for consumer groups using an _offset sync_ topic and _checkpoint_ topic.
+The _offset sync_ topic maps the source and target offsets for replicated topic partitions from record metadata.
+A _checkpoint_ is emitted from each source cluster and replicated in the target cluster through the _checkpoint_ topic.
+The _checkpoint_ topic maps the last committed offset in the source and target cluster for replicated topic partitions in each consumer group.
+
```
ajborley: It might be worth adding a para here about the fact that you need to add some code to your consumers if you want automatic fail-over - there is a new Kafka utility class (https://github.com/apache/kafka/blob/trunk/connect/mirror-client/src/main/java/org/apache/kafka/connect/mirror/RemoteClusterUtils.java) that you can use to translate your consumer group offset from the source cluster to the corresponding offset for the target cluster.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+        heartbeats.topic.replication.factor: 1
+    checkpointConnector:
+      config:
+        checkpoints.topic.replication.factor: 1
+```
+
+You can see the full `spec` options in the [KafkaMirrorMaker2 schema reference](https://strimzi.io/docs/master/#type-KafkaMirrorMaker2-reference).
+
+If, at this point, you're wondering what happens if you're using the old version of MirrorMaker.
+It's still supported, but you will need to updated to the format supported by MirrorMaker 2.0.
```
ajborley: Typo: `update` 
This line is a bit unclear - what do you mean by 'update to the format supported by MM2'? There are plans in Kafka to have a 'legacy' mode in MirrorMaker 2.0 (but it's not in Kafka 2.4.0) that will create topics without the cluster prefix and not do the topic config mirroring (which I think are the main behavioural changes between MM1 and 2).

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+        key: source.key
+        secretName: my-user-source
+      type: tls
+    bootstrapServers: my-cluster-source-kafka-bootstrap:9092
+    tls:
+      trustedCertificates:
+      - certificate: ca.crt
+        secretName: my-cluster-source-cluster-ca-cert
+  - alias: "my-cluster-target"
+    authentication: <10>
```
ajborley: I think this `<10>` is a copy & paste artifact.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+This image shows a single source cluster, but you can have multiple source clusters.
+
+![Connecting clusters with MirrorMaker 2.0](/assets/2020-03-12-mirrormaker.png)
+
+MirrorMaker 2.0 **_connectors_** -- remember, we're based on Kafka Connect now -- and related **_internal topics_** help manage the transfer and synchronization of data between the clusters.
+
+Different to the previous version of MirrorMaker, radically different, but different doesn't mean more complicated here.
+In fact, once you know the essentials, setting up is rather straightforward.
+
+Using Strimzi, you just need to work out how you want to configure a `KafkaMirrorMaker2` resource, which will define your MirrorMaker 2.0 deployment.
```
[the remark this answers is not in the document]
PaulRMellor: Good. Updated.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+Pretty good, right?
+
+This is MirrorMaker unleashed, ready to fulfill the potential we always knew it had.
+
+### Bidirectional opportunities
+
+In the previous version of MirrorMaker, the topic name in the source cluster is automatically created in the downstream cluster.
+Fine, to a point.
+But there is no distinction between _Topic-1.Partition-1_ in the source cluster and _Topic-1.Partition-1_ in the target cluster.
+Essentially, you are limited to _active/passive_ synchonization.
```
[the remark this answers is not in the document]
PaulRMellor: :+1:

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Fixed.

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Changed to:

> If, at this point, you're wondering what happens if you're using the old version of MirrorMaker,
> it's still supported.
> And there are also plans to have a _legacy mode_ in MirrorMaker 2.0 that creates topics without the cluster prefix, 
> and doesn't do the topic configuration mirroring.
> Basically, turning off the main differences between the original and the new version of MirrorMaker.

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Added:

> If you want automatic failover, you can add Kafka's new `RemoteClusterUtils.java` utility class to your consumers.
> The class translates translates the consumer group offset from the source cluster to the corresponding offset for the target cluster.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,247 @@
...
+A rose might be a rose by any other name, but when it comes to replicating topics in MirrorMaker 2.0 we can't afford such ambiguity.
+Particularly if we want to create a bidirectional configuration.
+
+The approach to topic renaming opens up a world of bidirectional opportunities.
+You can now have an _active/active_ cluster configuration that feeds data to and from each cluster.
+For this, you need a MirrorMaker 2.0 cluster at each destination.
+
+### Self-regulating topic replication
+
+Topic replication and consumer group replication is defined using regular expression patterns to _whitelist_ or _blacklist_ topics and consumer groups:
```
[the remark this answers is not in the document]
PaulRMellor: I've switched around as suggested. You might want to check it still makes sense.

PaulRMellor:
> Hi @PaulRMellor - I've added some comments & updates - is this the sort of input you wanted? It's looking good :)

Thanks @ajborley -- I've updated as suggested. You might want to look at how the content was rearranged for this comment:

> Consumer group replication is part of the MirrorCheckpointConnector, which hasn't been mentioned yet - perhaps just talk about topics here and add a paragraph to the Offset tracking and mapping that talks about the groupsPattern and groupsBlacklistPattern.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+Poof! These issues entirely disappear with MirrorMaker 2.0.
+Instead, we get simplicity and sophistication through the `MirrorCheckpointConnector`.
+
+`MirrorCheckpointConnector` tracks and maps offsets for specified consumer groups using an _offset sync_ topic and _checkpoint_ topic.
+The _offset sync_ topic maps the source and target offsets for replicated topic partitions from record metadata.
+A _checkpoint_ is emitted from each source cluster and replicated in the target cluster through the _checkpoint_ topic.
+The _checkpoint_ topic maps the last committed offset in the source and target cluster for replicated topic partitions in each consumer group.
+
+If you want automatic failover, you can add Kafka's new `RemoteClusterUtils.java` utility class to your consumers.
+The class translates translates the consumer group offset from the source cluster to the corresponding offset for the target cluster.
```
ajborley: typo: `translates translates` :)

PaulRMellor: Fixed :man_facepalming:

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+
+Poof! These issues entirely disappear with MirrorMaker 2.0.
+Instead, we get simplicity and sophistication through the `MirrorCheckpointConnector`.
+
+`MirrorCheckpointConnector` tracks and maps offsets for specified consumer groups using an _offset sync_ topic and _checkpoint_ topic.
+The _offset sync_ topic maps the source and target offsets for replicated topic partitions from record metadata.
+A _checkpoint_ is emitted from each source cluster and replicated in the target cluster through the _checkpoint_ topic.
+The _checkpoint_ topic maps the last committed offset in the source and target cluster for replicated topic partitions in each consumer group.
+
+If you want automatic failover, you can add Kafka's new `RemoteClusterUtils.java` utility class to your consumers.
```
scholzj: Technically, you use the `RemoteClusterUtils` class. But you do not add it by adding the clas to the project. You instead add it by adding the `connect-mirror-client` as dependency. 

```xml
        <dependency>
            <groupId>org.apache.kafka</groupId>
            <artifactId>connect-mirror-client</artifactId>
            <version>2.4.0</version>
        </dependency>
```

Maybe we do not need to show the XML, but I think we need to make this a bit more technically correct.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+The consumer groups tracked by `MirrorCheckpointConnector` are dependent on those defined in a _whitelist_ or _blacklist_:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1alpha1
+kind: KafkaMirrorMaker2
+metadata:
+  name: my-mirror-maker2
+spec:
+  # ...
+  groupsPattern: "group1|group2|group3"
```
scholzj: Again, from the example, this structure is a bit more complicated and we should make this match it.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+      config.storage.replication.factor: 1
+      offset.storage.replication.factor: 1
+      status.storage.replication.factor: 1
+    tls:
+      trustedCertificates:
+      - certificate: ca.crt
+        secretName: my-cluster-target-cluster-ca-cert
+```
+
+People can be confused by the idea of _replicas_ and _replication_ through MirrorMaker,
```
scholzj: Given you talked abotu replicas just above the example as Kubernetes replcias, do we need to clarify that here as well?

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+      offset-syncs.topic.replication.factor: 1
+  heartbeatConnector:
+    config:
+      heartbeats.topic.replication.factor: 1
+  checkpointConnector:
+    config:
+      checkpoints.topic.replication.factor: 1
+```
+
+You can see the full `spec` options in the [KafkaMirrorMaker2 schema reference](https://strimzi.io/docs/master/#type-KafkaMirrorMaker2-reference).
```
scholzj:
```suggestion
You can see the full `spec` options in the [KafkaMirrorMaker2 schema reference](https://strimzi.io/docs/latest/#type-KafkaMirrorMaker2-reference).
```

I think we should point here to latest or to 0.17.0 docs. Not to master. (when we publish it, the link should work in latest as well.)

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+
+In the previous version of MirrorMaker, the topic name in the source cluster is automatically created in the downstream cluster.
+Fine, to a point.
+But there is no distinction between _Topic-1.Partition-1_ in the source cluster and _Topic-1.Partition-1_ in the target cluster.
+Essentially, you are limited to _active/passive_ synchronization.
+MirrorMaker 2.0 solves this by introducing the concept of _remote_ topics.
+
+_Remote_ topics are created from the originating cluster by the `MirrorSourceConnector`.
+They are distinguished by automatic renaming that appends the name of cluster to the name of the topic.
+Our _Topic-1.Partition-1_ from the source cluster becomes the never-to-be-confused _Cluster-1-Topic-1.Partition-1_ in the target cluster.
```
ppatierno: it _Cluster-1.Topic-1.Partition-1_ (see the "." after the cluster name instead of the "-")

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+And what about **_topic configuration synchronization_**  and **_offset tracking_**?
+Pretty good, right?
+
+This is MirrorMaker unleashed, ready to fulfill the potential we always knew it had.
+
+### Bidirectional opportunities
+
+In the previous version of MirrorMaker, the topic name in the source cluster is automatically created in the downstream cluster.
+Fine, to a point.
+But there is no distinction between _Topic-1.Partition-1_ in the source cluster and _Topic-1.Partition-1_ in the target cluster.
```
ppatierno: I am not convinced that we should show the partition in the example.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+You use `groupsBlacklistPattern` if you want to use blacklists.
+
+### Checking the connection
+
+The old way of checking that MirrorMaker was working was by using standard Kubernetes _healthcheck_ probes to know when MirrorMaker can accept traffic and when it needs a restart.
+MirrorMaker 2.0 periodically checks on the connection through its dedicated `MirrorHeartbeatConnector`.
+
+`MirrorHeartbeatConnector` periodically checks connectivity between clusters using the _heartbeat_ topic.
+A _heartbeat_ is emitted from each source cluster and replicated in the target cluster through the _heartbeat_ topic.
+The _heartbeat_ topic is used to check that the source cluster is available and the clusters are connected.
```
ppatierno: I have a more technical question here because the paragraph is not clear to me.
If I have just a source and a target cluster, MM2 is running on the target one. It's based on Kafka Connect and works in pull mode, pulling messages from the source; it sounds to me that the source cluster doesn't know that "someone" is mirroring the messages, it doesn't know about MM2 on the other side. So how can it produce a heartbeat message into a local heartbeat topic? Tbh it seems to be what the KIP-382 says (https://cwiki.apache.org/confluence/display/KAFKA/KIP-382%3A+MirrorMaker+2.0) but in one of my deployments I just see a `heartbeats` topic on the target cluster where MM2 is running and nothing related to heartbeat on the source cluster. @ajborley  maybe a question for you?

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+kind: KafkaMirrorMaker2
+metadata:
+  name: my-mirror-maker2
+spec:
+  version: {DefaultKafkaVersion}
+  replicas: 3
+  connectCluster: "my-cluster-target"
+    clusters:
+    - alias: "my-cluster-source"
+    authentication:
```
ppatierno: wrong indentation

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+      certificateAndKey:
+        certificate: source.crt
+        key: source.key
+        secretName: my-user-source
+      type: tls
+    bootstrapServers: my-cluster-source-kafka-bootstrap:9092
+    tls:
```
ppatierno: as above, wrong indentation

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+
+If, at this point, you're wondering what happens if you're using the old version of MirrorMaker,
+it's still supported.
+And there are also plans to have a _legacy mode_ in MirrorMaker 2.0 that creates topics without the cluster prefix,
+and doesn't do the topic configuration mirroring.
+Basically, turning off the main differences between the original and the new version of MirrorMaker.
+
+### Embrace change
+
+Apache understood that MirrorMaker was due an overhaul [[KIP-382: MirrorMaker 2.0](https://cwiki.apache.org/confluence/display/KAFKA/KIP-382%3A+MirrorMaker+2.0)].
```
ppatierno: I would not say "Apache" here but the "Apache Kafka community"

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+This time the topic configuration is automatically synchronized between source and target clusters through configuration properties.
+Configuration changes are propagated to remote topics so that new topics and partitions are detected and created.
+The need for rebalancing vanishes.
+
+### Offset tracking and mapping
+
+In the old version of MirrorMaker, the offset of the source topic in the target cluster begins when the replication begins.
+The `__consumer_offsets` topic is not mirrored.
+So offsets of the source topic and its replicated equivalent can have two entirely different positions.
+This can be problematic in a failover situation.
```
tombentley:
```suggestion
This was often problematic in a failover situation.
```

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+      status.storage.replication.factor: 1
+    tls:
+      trustedCertificates:
+      - certificate: ca.crt
+        secretName: my-cluster-target-cluster-ca-cert
+```
+
+People can be confused by the idea of _replicas_ and _replication_ through MirrorMaker,
+that's why you'll see replication between clusters often referred to as _mirroring_.
+Just thought I'd bring that up in case I've been inconsistent in this post.
```
tombentley: I think we need to be consistent in this post, rather than explain it part way through with an apology.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+
+If you don't want to leave the defaults, you can also include configuration for the **_MirrorMaker 2.0 connectors_** and related **_internal topics_**.
```
tombentley: Do we need both bold and italic?

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+If, at this point, you're wondering what happens if you're using the old version of MirrorMaker,
+it's still supported.
+And there are also plans to have a _legacy mode_ in MirrorMaker 2.0 that creates topics without the cluster prefix,
+and doesn't do the topic configuration mirroring.
+Basically, turning off the main differences between the original and the new version of MirrorMaker.
+
+### Embrace change
+
+Apache understood that MirrorMaker was due an overhaul [[KIP-382: MirrorMaker 2.0](https://cwiki.apache.org/confluence/display/KAFKA/KIP-382%3A+MirrorMaker+2.0)].
+They identified the key issues with using the original MirrorMaker -- manual topic configuration, the lack of support for _active/active_ replication, and the inability to track offsets -- and eradicated them with MirrorMaker 2.0.
```
tombentley: Can we reword this to avoid "they", that kind of supposes some kind of "them" and "us" split which is not the case. "Key usage issues with the original MorrirMaker... were identified" for example.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+title:  "Introducing MirrorMaker 2.0 to Strimzi"
+date: 2020-03-12
+author: paul_mellor
+---
+
+Apache Kafka MirrorMaker replicates data between two Kafka clusters, within or across data centers.
+MirrorMaker takes messages from a source Kafka cluster and writes them to a target Kafka cluster,
+which makes it a very useful tool for those wanting to ensure the availability and consistency of their enterprise data.
+And who doesn't?
+Typical scenarios where you might consider MirrorMaker are for disaster recovery and increased throughput.
```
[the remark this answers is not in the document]
PaulRMellor: I've changed to:

_ Typical scenarios where you might consider MirrorMaker are for disaster recovery and data aggregation._

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+
+With the release of Strimzi 0.17, things become a little more interesting with the introduction of support for MirrorMaker 2.0.
+MirrorMaker 2.0 represents a significant shift in the way you synchronize data between replicated Kafka clusters,
+promising a more dynamic and automated approach to topic replication between clusters.
```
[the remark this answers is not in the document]
PaulRMellor: I'm hoping they read on to find out. :-) We can introduce the concepts of topic config sync and offset tracking here, but we'll be explaining it later?

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+Topic replication is defined using regular expression patterns to _whitelist_ or _blacklist_ topics:
+
+```yaml
+apiVersion: kafka.strimzi.io/v1alpha1
+kind: KafkaMirrorMaker2
+metadata:
+  name: my-mirror-maker2
+spec:
+  # ...
+  topicsPattern: ".*"
```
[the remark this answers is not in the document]
PaulRMellor: We're only talking about the patterns here.

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: I've changed to:

> If you want automatic failover, you can use Kafka's new `RemoteClusterUtils.java` utility class by adding `connect-mirror-client` as a dependency to your consumers.
> 
> ```xml
> <dependency>
>     <groupId>org.apache.kafka</groupId>
>     <artifactId>connect-mirror-client</artifactId>
>     <version>2.4.0</version>
> </dependency>
> ```

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Changed to:

> _People can be confused by the idea of _replicas_ and _replication_ through MirrorMaker,
> that's why you'll see replication between clusters often referred to as _mirroring_ so that it's not confused with 
> the `replicas` that represent the nodes replicated in a deployment._

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Changed to _latest_

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,268 @@
...
+
+The previous version of MirrorMaker relies on configuration of a _source producer_ and _target consumer_ pair to synchronize Kafka clusters.
+
+MirrorMaker 2.0 is based on Kafka Connect, which is a bit of a game changer.
+For a start, there is no longer a need to configure producers and consumers to make your connection.
+
+With MirrorMaker 2.0 deployed with Strimzi, you identify your source and target clusters.
+You then configure and deploy MirrorMaker 2.0 to make the connection between those clusters.
+
+This image shows a single source cluster, but you can have multiple source clusters.
```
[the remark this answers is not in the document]
PaulRMellor: Great idea! Added.

> _This image shows a single source cluster, but you can have multiple source clusters.
> Something that was not possible with old MirrorMaker._
>

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: We use _Cluster-1-_ in the image too

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: It's the same in the guide.

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Thanks Paolo. @ajborley - heartbeats and source clusters. Is there anything we can add to clarify what happens to produce a heartbeat at the source?

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Excellent

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Yes. I've removed that. :+1:

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: I guess not. I've left just for the features we want to highlight.

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Good idea. Changed.
Paolo mentioned using _Apache Kafka community_ as well.

> _Key issues with using the original MirrorMaker were identified -- manual topic configuration, the lack of support for _active/active_ replication, and the inability to track offsets -- and eradicated with MirrorMaker 2.0._

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: I'm not sure I'm getting what you want to change here. We're just talking about patterns and the examples you sent in the link show:

>  topicsPattern: ".*"
>  groupsPattern: ".*"

_posts/2020-03-06-introducing-mirrormaker2.md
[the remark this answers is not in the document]
scholzj: Ok, fair enough.

_posts/2020-03-06-introducing-mirrormaker2.md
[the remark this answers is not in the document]
scholzj: Yes, but this will confuse the users. After reading this I would think that this is on the level directly under `spec`. In all the examples in the docs we always follow the whole structure. So I think it should be something like this:

```yaml
apiVersion: kafka.strimzi.io/v1alpha1
kind: KafkaMirrorMaker2
metadata:
  name: my-mirror-maker2
spec:
  # ...
  mirrors:
    - sourceCluster: ...
      targetCluster: ...
      topicsPattern: ".*"
      # ...
```

_posts/2020-03-06-introducing-mirrormaker2.md
scholzj: Yes, but this will confuse the users. After reading this I would think that this is on the level directly under `spec`. In all the examples in the docs we always follow the whole structure. So I think it should be something like this:

```yaml
apiVersion: kafka.strimzi.io/v1alpha1
kind: KafkaMirrorMaker2
metadata:
  name: my-mirror-maker2
spec:
  # ...
  mirrors:
    - sourceCluster: ...
      targetCluster: ...
      groupsPattern: ".*"
      # ...
```

_posts/2020-03-06-introducing-mirrormaker2.md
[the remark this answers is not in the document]
PaulRMellor: Thanks. I see what you're saying. Updated.

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Got it. Updated.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+### Bidirectional opportunities
+
+In the previous version of MirrorMaker, the topic name in the source cluster is automatically created in the downstream cluster.
+Fine, to a point.
+But there is no distinction between _Topic-1.Partition-1_ in the source cluster and _Topic-1.Partition-1_ in the target cluster.
+Essentially, you are limited to _active/passive_ synchronization.
+MirrorMaker 2.0 solves this by introducing the concept of _remote_ topics.
+
+_Remote_ topics are created from the originating cluster by the `MirrorSourceConnector`.
+They are distinguished by automatic renaming that appends the name of cluster to the name of the topic.
```
tombentley:
```suggestion
They are distinguished by automatic renaming that prepends the name of cluster to the name of the topic.
```

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+Our _Topic-1.Partition-1_ from the source cluster becomes the never-to-be-confused _Cluster-1-Topic-1.Partition-1_ in the target cluster.
```
tombentley: It's not simply that the name is unambigous. It's also that a consumer in the target cluster can now consume _both_ _Topic-1_ and _Cluster-1-Topic-1_ while still having unambigous Kafka-side consumer offsets.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+As you can see here, _remote_ topics can be easily identified so that they are not returned to the source cluster.
```
tombentley: "not returned to the source cluster" would be clearer to me if you said something like "...so that we can ensure there are not cycles in the graph of replication".

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+Why _remote_ topic and not just _source_ topic?
+Because each source cluster can also be a target cluster in a bidirectional configuration.
+This feature also means that it is no longer necessary to aggregate data in a separate cluster.
+
+A rose might be a rose by any other name, but when it comes to replicating topics in MirrorMaker 2.0 we can't afford such ambiguity.
+Particularly if we want to create a bidirectional configuration.
+
+The approach to topic renaming opens up a world of bidirectional opportunities.
+You can now have an _active/active_ cluster configuration that feeds data to and from each cluster.
+For this, you need a MirrorMaker 2.0 cluster at each destination.
```
tombentley: I think you need a diagram or a bit more explanation about how this works.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+  mirrors:
+  - sourceCluster: "my-cluster-source"
+    targetCluster: "my-cluster-target"
+    topicsPattern: ".*"
+    # ...
+```
+
+You use `topicsBlacklistPattern` if you want to use blacklists.
+
+Pretty similar to before, but previously topic _configuration_ was not replicated.
```
tombentley:
```suggestion
This is pretty similar to MirrorMaker1, but previously topic _configuration_ was not replicated.
```

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+The target cluster had to be manually maintained to match topic configuration changes in the source cluster.
```
tombentley:
```suggestion
The target cluster had to be manually maintained to match topic configuration changes in the source cluster.
Something easily overlooked, and typically requiring users to build their own automation to achieve at scale.
```

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+This time the topic configuration is automatically synchronized between source and target clusters through configuration properties.
```
tombentley:
```suggestion
With MM2 the topic configuration is automatically synchronized between source and target clusters through configuration properties.
```

But what are "configuration properties". Does it mean "this can be done by configuring it in the MM2 CR"? If that's the configuration you mean then say that.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+Configuration changes are propagated to remote topics so that new topics and partitions are detected and created.
+The need for rebalancing vanishes.
```
tombentley: Where did rebalancing come from? Rebalancing of what?

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Offset tracking and mapping
+
+In the old version of MirrorMaker, the offset of the source topic in the target cluster begins when the replication begins.
```
tombentley:
```suggestion
In MirrorMaker 1, the offset of the source topic in the target cluster begins when the replication begins.
```

But I don't understand what this means. The source topic doesn't itself have an offset. A consumer of the source topic has an offset, but the sentence doesn't mention consumers.

_posts/2020-03-06-introducing-mirrormaker2.md
ajborley: The heartbeat topic and records are created on the target cluster (the cluster that is local to MM2/Kafka Connect), not the source cluster (remote). The idea is to have MM2 clusters on both sides, so that a heartbeat produced locally by the HeartbeatConnector running at the source cluster is treated as a normal remote topic to be mirrored by the MM2 SourceConnector running at the target.
 
By mirroring the heartbeat topics, it’s easy to check that the connection between source and target is stable and if things do go wrong, the offset position and record time stamps can help with recovery and diagnosis.

Hope that makes some kind of sense!

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+The `__consumer_offsets` topic is not mirrored.
+So offsets of the source topic and its replicated equivalent can have two entirely different positions.
+This was often problematic in a failover situation.
+How to find the offset in the target cluster?
+Strategies such as using timestamps can be adopted, but it adds complexity.
+
+Poof! These issues entirely disappear with MirrorMaker 2.0.
```
tombentley:
```suggestion
These issues entirely disappear with MirrorMaker 2.0.
```

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+    targetCluster: "my-cluster-target"
+    groupsPattern: "group1|group2|group3"
+    # ...
+```
+
+You use `groupsBlacklistPattern` if you want to use blacklists.
+
+### Checking the connection
+
+The old way of checking that MirrorMaker was working was by using standard Kubernetes _healthcheck_ probes to know when MirrorMaker can accept traffic and when it needs a restart.
```
tombentley:
```suggestion
The old way of checking that MirrorMaker was working, on Kubernetes at least, was by using standard Kubernetes _healthcheck_ probes to know when MirrorMaker can accept traffic and when it needs a restart.
```

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Updated to:

> `MirrorHeartbeatConnector` periodically checks connectivity between clusters using the _heartbeat_ topic.
> A _heartbeat_ is emitted from each remote (source) cluster and mirrored to the target cluster by the `MirrorSourceConnector` as a _heartbeat_ topic.
> The _heartbeat_ topic makes it easy to check that the source cluster is available and the clusters are connected.
> If things go wrong, the _heartbeat_ topic offset positions and time stamps can help with recovery and diagnosis.

ppatierno: @PaulRMellor but it doesn't explain the scenario where you have just one MM2 on the target cluster. So we should be clear that what you wrote makes sense when you have MM2 on both source and target otherwise it doesn't. 

@ajborley reading your explanation (thanks for that!) sounds to me that when you have a scenario with MM2 just running on the target, the _hearthbeat_ topic is quite useless, is that right?

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+With the release of Strimzi 0.17, things become a little more interesting with the introduction of support for MirrorMaker 2.0.
+MirrorMaker 2.0 represents a significant shift in the way you synchronize data between replicated Kafka clusters,
+promising a more dynamic and automated approach to topic replication between clusters.
+
+<!--more-->
+
+### MirrorMaker 2.0 - the Kafka Connect(ion)
+
+The previous version of MirrorMaker relies on configuration of a _source producer_ and _target consumer_ pair to synchronize Kafka clusters.
```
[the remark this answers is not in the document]
PaulRMellor: :open_mouth:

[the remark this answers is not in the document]
PaulRMellor: Fixed

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+With MirrorMaker 2.0 deployed with Strimzi, you identify your source and target clusters.
+You then configure and deploy MirrorMaker 2.0 to make the connection between those clusters.
+
+This image shows a single source cluster, but you can have multiple source clusters.
+Something that was not possible with old MirrorMaker.
+
+![Connecting clusters with MirrorMaker 2.0](/assets/2020-03-12-mirrormaker.png)
+
+MirrorMaker 2.0 _connectors_ -- remember, we're based on Kafka Connect now -- and related _internal topics_ help manage the transfer and synchronization of data between the clusters.
```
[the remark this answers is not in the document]
PaulRMellor: I've mentioned them in brackets

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+In fact, once you know the essentials, setting up is rather straightforward.
+
+Using Strimzi, you just need to work out how you want to configure a `KafkaMirrorMaker2` resource, which will define your Kafka Connect deployment and start running a set of MirrorMaker 2.0 connectors.
+We'll look a bit more at configuring the `KafkaMirrorMaker2` resource later on.
+
+First, something you'll be keen to know.
+Is the sequel better than the original?
+Well, how does **_bidirectional replication_** sound?
+And what about **_topic configuration synchronization_**  and **_offset tracking_**?
+Pretty good, right?
```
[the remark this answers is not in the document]
PaulRMellor: I've moved it to the top of this section

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+This is MirrorMaker unleashed, ready to fulfill the potential we always knew it had.
```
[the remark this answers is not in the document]
PaulRMellor: I've toned it down a bit:

> This is MirrorMaker ready to fulfill the potential we always knew it had.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Bidirectional opportunities
+
+In the previous version of MirrorMaker, the topic name in the source cluster is automatically created in the downstream cluster.
+Fine, to a point.
+But there is no distinction between _Topic-1.Partition-1_ in the source cluster and _Topic-1.Partition-1_ in the target cluster.
+Essentially, you are limited to _active/passive_ synchronization.
```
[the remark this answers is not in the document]
PaulRMellor: I've added: _, because _active/active_ replication would cause an infinite loop._

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Thanks. I've added:

> The target cluster can consume _Topic-1.Partition-1_ and  _Cluster-1-Topic-1.Partition-1_, and have unambiguous consumer offsets.

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: This makes it less clear to me. :disappointed:  Do we need to go into such technicalities for this blog post?

tombentley: How about "... so there's no possibility of messages being sent back and forth in a loop"

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: The previous images shows it, so I've added: _, as shown in the previous image._

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: I changed to:

> _With MirrorMaker 2.0, the topic configuration is automatically synchronized between source and target clusters according to the topics defined in the `MirrorMaker2` custom resource._

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: I changed to:

> _The need for consumer rebalancing due to topic changes is greatly reduced._

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Changed to:

> _the consumer offset of the source topic in the target cluster begins when the replication begins._

_posts/2020-03-06-introducing-mirrormaker2.md
PaulRMellor: Not Poof! That was my favourite word in the post. :disappointed:  Removed.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+![Connecting clusters with MirrorMaker 2.0](/assets/2020-03-12-mirrormaker.png)
+
+MirrorMaker 2.0 _connectors_ -- remember, we're based on Kafka Connect now -- and related _internal topics_ help manage the transfer and synchronization of data between the clusters.
+
+Different to the previous version of MirrorMaker, radically different, but different doesn't mean more complicated here.
+In fact, once you know the essentials, setting up is rather straightforward.
+
+Using Strimzi, you just need to work out how you want to configure a `KafkaMirrorMaker2` resource, which will define your Kafka Connect deployment and start running a set of MirrorMaker 2.0 connectors.
+We'll look a bit more at configuring the `KafkaMirrorMaker2` resource later on.
```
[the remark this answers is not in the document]
PaulRMellor: That's how you keep people reading. :wink:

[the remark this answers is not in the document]
PaulRMellor: I've removed this line now.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+This image shows a single source cluster, but you can have multiple source clusters.
+Something that was not possible with old MirrorMaker.
+
+![Connecting clusters with MirrorMaker 2.0](/assets/2020-03-12-mirrormaker.png)
+
+MirrorMaker 2.0 _connectors_ -- remember, we're based on Kafka Connect now -- and related _internal topics_ help manage the transfer and synchronization of data between the clusters.
+
+Different to the previous version of MirrorMaker, radically different, but different doesn't mean more complicated here.
+In fact, once you know the essentials, setting up is rather straightforward.
```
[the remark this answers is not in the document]
PaulRMellor: This is the intro. They read on to find out. It's kind of a review up to now.

_posts/2020-03-06-introducing-mirrormaker2.md
```diff
@@ -0,0 +1,285 @@
...
+
+Using Strimzi, you just need to work out how you want to configure a `KafkaMirrorMaker2` resource, which will define your Kafka Connect deployment and start running a set of MirrorMaker 2.0 connectors.
```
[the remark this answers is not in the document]
PaulRMellor: Changed to:

_Using Strimzi, you configure a `KafkaMirrorMaker2` resource to define the Kafka Connect deployment, including the connection details of the source and target clusters, and start running a set of MirrorMaker 2.0 connectors to make the connection._

PaulRMellor:
@tombentley - Thanks Tom. Great comments. Let me know if I've misinterpreted any.

_posts/2020-03-06-introducing-mirrormaker2.md
ajborley: @ppatierno - agreed, the heartbeat topic doesn’t have a lot of utility when it’s not being mirrored. It does show you that your MM2 cluster can continuously connect and produce to the local Kafka, which could perhaps be useful, particularly when the topics that are being mirrored to the local Kafka (via the MM2 MirrorSourceConnector) are bursty or sporadic.

ppatierno: @ajborley agreed. So @PaulRMellor I think we should reflect this much better in the above paragraph.

PaulRMellor: How about this:

> `MirrorHeartbeatConnector` periodically checks connectivity between clusters.
> A _heartbeat_ is emitted from each remote (source) cluster and a _heartbeat_ topic is created on a target cluster. 
> If you have MirrorMaker 2.0 at both the remote and target location, a _heartbeat_ emitted locally by the `MirrorHeartbeatConnector` is treated like any remote topic and mirrored by the `MirrorSourceConnector` at the target cluster.
> The _heartbeat_ topic makes it easy to check that the source cluster is available and the clusters are connected.
> If things go wrong, the _heartbeat_ topic offset positions and time stamps can help with recovery and diagnosis.

ajborley: > A _heartbeat_ is emitted from each remote (source) cluster and a heartbeat topic is created on a target cluster.

This isn't quite right - how about:

 A _heartbeat_ record is produced every second by the `MirrorHeartbeatConnector` into a _heartbeat_ topic that is created on the target cluster.

ppatierno: Well, source and target clusters naming starting to be not well defined.
For example, consider the case where I have a source cluster (where producers send messages) and a target cluster (where consumers get mirrored messages). I have MM2 on the target cluster for mirroring but we decide to leverage the heartbeat feature so we need to install MM2 on the source cluster as well. In any case, the source cluster doesn't want to mirror any other topics from the target but just the heartbeat right? In this case, a heartbeat topic is created on the source cluster as well and mirrored to the target, isn't it?

PaulRMellor: Thanks @ajborley. Latest:

> `MirrorHeartbeatConnector` periodically checks connectivity between clusters.
> A _heartbeat_ is produced every second by the `MirrorHeartbeatConnector` into a _heartbeat_ topic that is created on the target cluster.
> If you have MirrorMaker 2.0 at both the remote and target location, the _heartbeat_ emitted locally by the `MirrorHeartbeatConnector` is treated like any remote topic and mirrored by the `MirrorSourceConnector` at the target cluster.
> The _heartbeat_ topic makes it easy to check that the source cluster is available and the clusters are connected.
> If things go wrong, the _heartbeat_ topic offset positions and time stamps can help with recovery and diagnosis.

ajborley: @ppatierno - you're right. In your case the MM2 running at the source cluster is only running the `MirrorHeartbeatConnector`, and the MM2 running at the target cluster is only running the `MirrorSourceConnector`, causing the `heartbeat` topic to be created on the source cluster and mirrored to the target cluster.

Perhaps we should change this to use the terms `local` and `remote`? How does the following sound:

> `MirrorHeartbeatConnector` periodically checks connectivity between clusters.
A _heartbeat_ is produced every second by the `MirrorHeartbeatConnector` into a _heartbeat_ topic that is created on the local cluster.
If you have MirrorMaker 2.0 at both the remote and local locations, the _heartbeat_ emitted at the remote location by the `MirrorHeartbeatConnector` is treated like any remote topic and mirrored by the `MirrorSourceConnector` at the local cluster.
The _heartbeat_ topic makes it easy to check that the remote cluster is available and the clusters are connected.
If things go wrong, the _heartbeat_ topic offset positions and time stamps can help with recovery and diagnosis.

ppatierno: @ajborley yeah I like it. @PaulRMellor wdyt?

PaulRMellor: Good for me :+1:

PaulRMellor: @ppatierno@ajborley. Thanks. I've updated as suggested.

ppatierno: we are good to merge than!

PR #126, opened 2020-03-18
Add Kubernetes Kind instructions

mstruk:
Signed-off-by: Marko Strukelj <marko.strukelj@gmail.com>

### DIFF
--- a/quickstarts/kind/index.md
+++ b/quickstarts/kind/index.md
@@ -3,13 +3,13 @@ title: Kubernes Kind Quickstart
 layout: default
 ---
 
-Kubernetes Kind is a single-node Kubernetes cluster implemented as a single Docker image that runs as a container.
+Kubernetes Kind is a Kubernetes cluster implemented as a single Docker image that runs as a container.
 It was primarily designed for testing Kubernetes itself, but may be used for local development or CI.
 
 When using a local install of Minikube or Minishift, the Kubernetes cluster is started inside a virtual machine, running a Linux kernel and a Docker daemon, consuming extra CPU and RAM.
 
 Kind, on the other hand, requires no additional VM - it simply runs as a linux container with a set of processes using the same Linux kernel used by your Docker daemon.
-For this reason it is faster to start, and consumes less CPU and RAM than the alternatives.
+For this reason it is faster to start, and consumes less CPU and RAM than the alternatives, which is especially noticeable when running a Docker daemon natively on a Linux host.
 
 # Installing the dependencies
 
@@ -81,7 +81,7 @@ kubectl logs `kubectl get pod -n kafka -o name | grep strimzi-cluster-operator` 
 
 # Provision the Apache Kafka cluster
 
-After that we feed Strimzi with a simple **Custom Resource**, which will then give you a small persistent Apache Kafka Cluster with one node for each, Apache Zookeeper and Apache Kafka:
+Then we create a new Kafka custom resource, which will give us a small persistent Apache Kafka Cluster with one node for each - Apache Zookeeper and Apache Kafka:
 
 ```shell
 # Apply the `Kafka` Cluster CR file
--- a/quickstarts/kind/index.md
+++ b/quickstarts/kind/index.md
@@ -1,0 +1,107 @@
+---
+title: Kubernes Kind Quickstart
+layout: default
+---
+
+# Installing the dependencies
+
+This quickstart assumes that you have the latest version of the `kind` binary, which you can get [here](https://github.com/kubernetes-sigs/kind/releases).
+
+Kind requires a running Docker Daemon. There are different Docker options depending on your host platform.
+You can follow the instructions [here](https://docs.docker.com/get-docker/).
+
+You'll also need `kubectl` binary, which you can get by following the instruction [here](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
+
+Once you have all the binaries installed, and a Docker daemon running, make sure everything works:
+
+```shell
+# Validate docker installation
+docker ps
+docker version
+
+# Validate kind
+kind version
+
+# Validate kubectl
+kubectl version
+```
+
+# Configuring the Docker daemon
+
+If your Docker Daemon runs as a VM you'll most likely need to configure how much memory the VM should have, how many CPUs, how much disk space, and swap size.
+Make sure to assign at least 2 CPUs, and preferably 4 Gb or more of RAM. Consult Docker documentation for you platform.
+
+
+# Starting Kubernetes cluster
+
+This will start a local development cluster of [Kubernetes Kind](https://github.com/kubernetes-sigs/kind) which installs as a single docker container.
+
+```shell
+# Install kind
+kind create cluster
+```
+
+# Applying Strimzi installation file
+
+Before deploying Strimzi Kafka operator, let's first create our `kafka` namespace:
+
+```shell
+kubectl create namespace kafka
+```
+
+Next we apply the Strimzi install files, including `ClusterRoles`, `ClusterRoleBindings` and some **Custom Resource Definitions** (`CRDs`). The CRDs define the schemas used for declarative management of the Kafka cluster, Kafka topics and users.
+
+```shell
+curl -L https://github.com/strimzi/strimzi-kafka-operator/releases/download/{{site.data.releases.operator[0].version}}/strimzi-cluster-operator-{{site.data.releases.operator[0].version}}.yaml \
+  | sed 's/namespace: .*/namespace: kafka/' \
+  | kubectl apply -f - -n kafka 
+```
+
+Note how we set all the `namespace` references in downloaded .yaml file to `kafka`. By default they are set to `myproject`.
+But we want them all to be `kafka` because we decided to install the operator into `kafka` namespace, which we achieve by specifying `-n kafka` when running `kubectl apply` ensuring that all the definitions and the configurations are installed in `kafka` namespace rather than the `default` namespace.
+
+If there is a mismatch between namespaces, then the Strimzi Cluster Operator will not have the necessary permissions to perform its operations. 
+
+Follow the deployment of strimzi kafka operator:
+```shell
+kubectl get pod -n kafka --watch
+```
+
+You can also follow the operator's log:
+```shell
+kubectl logs `kubectl get pod -n kafka | grep strimzi-cluster-operator | awk '{print $1}'` -n kafka -f
+```
+
+
+# Provision the Apache Kafka cluster
+
+After that we feed Strimzi with a simple **Custom Resource**, which will then give you a small persistent Apache Kafka Cluster with one node for each, Apache Zookeeper and Apache Kafka:
+
+```shell
+# Apply the `Kafka` Cluster CR file
+kubectl apply -f https://raw.githubusercontent.com/strimzi/strimzi-kafka-operator/{{site.data.releases.operator[0].version}}/examples/kafka/kafka-persistent-single.yaml -n kafka 
+```
+
+We now need to wait while Kubernetes starts the required pods, services and so on:
+
+```shell
+kubectl wait kafka/my-cluster --for=condition=Ready --timeout=300s -n kafka 
+```
+
+The above command might timeout if you're downloading images over a slow connection. If that happens you can always run it again.
+
+# Send and receive messages
+
+Once the cluster is running, you can run a simple producer to send messages to a Kafka topic (the topic will be automatically created):
+
+```shell
+kubectl -n kafka run kafka-producer -ti --image=strimzi/kafka:{{site.data.releases.operator[0].version}}-kafka-{{site.data.releases.operator[0].defaultKafkaVersion}} --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list my-cluster-kafka-bootstrap:9092 --topic my-topic
+```
+
+And to receive them in a different terminal you can run:
+
+```shell
+kubectl -n kafka run kafka-consumer -ti --image=strimzi/kafka:{{site.data.releases.operator[0].version}}-kafka-{{site.data.releases.operator[0].defaultKafkaVersion}} --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --topic my-topic --from-beginning
+```
+
+Enjoy your Apache Kafka cluster, running on Minikube!
--- a/quickstarts/kind/index.md
+++ b/quickstarts/kind/index.md
@@ -37,7 +37,7 @@ kubectl version
 # Configuring the Docker daemon
 
 If your Docker Daemon runs as a VM you'll most likely need to configure how much memory the VM should have, how many CPUs, how much disk space, and swap size.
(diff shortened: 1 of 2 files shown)


quickstarts/kind/index.md
```diff
@@ -0,0 +1,107 @@
...
+kubectl -n kafka run kafka-producer -ti --image=strimzi/kafka:{{site.data.releases.operator[0].version}}-kafka-{{site.data.releases.operator[0].defaultKafkaVersion}} --rm=true --restart=Never -- bin/kafka-console-producer.sh --broker-list my-cluster-kafka-bootstrap:9092 --topic my-topic
+```
+
+And to receive them in a different terminal you can run:
+
+```shell
+kubectl -n kafka run kafka-consumer -ti --image=strimzi/kafka:{{site.data.releases.operator[0].version}}-kafka-{{site.data.releases.operator[0].defaultKafkaVersion}} --rm=true --restart=Never -- bin/kafka-console-consumer.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --topic my-topic --from-beginning
+```
+
+Enjoy your Apache Kafka cluster, running on Minikube!
```
scholzj:
```suggestion
Enjoy your Apache Kafka cluster, running on Kind!
```

quickstarts/kind/index.md
```diff
@@ -0,0 +1,107 @@
...
+Follow the deployment of strimzi kafka operator:
+```shell
+kubectl get pod -n kafka --watch
+```
+
+You can also follow the operator's log:
+```shell
+kubectl logs `kubectl get pod -n kafka | grep strimzi-cluster-operator | awk '{print $1}'` -n kafka -f
+```
+
```
scholzj: Double empty line not needed I guess.

quickstarts/kind/index.md
```diff
@@ -0,0 +1,107 @@
...
+If your Docker Daemon runs as a VM you'll most likely need to configure how much memory the VM should have, how many CPUs, how much disk space, and swap size.
+Make sure to assign at least 2 CPUs, and preferably 4 Gb or more of RAM. Consult Docker documentation for you platform.
+
+
+# Starting Kubernetes cluster
+
+This will start a local development cluster of [Kubernetes Kind](https://github.com/kubernetes-sigs/kind) which installs as a single docker container.
+
+```shell
+# Install kind
```
scholzj: I wonder if we need this comment really. We do not have it in the follow up commands. But in any case, if you keep it it should probably say _Create Kind cluster_ nd not _Install kind_?

mstruk: True. Will remove it.

quickstarts/kind/index.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Kubernes Kind Quickstart
+layout: default
+---
+
+# Installing the dependencies
```
ppatierno: Maybe we could add a little bit of introduction about Kind? Even providing a comparison with Minikube?

mstruk: I haven't added it simply because there is no such introduction for Minikube or Minishift either, and I just followed the same template. But I can add some ...

scholzj: I would not get into too many details to be honest. Too muhc stuff to maintain which everyone can learn from the linked pages.

mstruk: I hope it's not too much: https://github.com/strimzi/strimzi.github.io/pull/126/commits/7731702c59467c5689f6fc780e98f9f08804565b

ppatierno: +1

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
+---
+title: Kubernes Kind Quickstart
+layout: default
+---
+
+Kubernetes Kind is a single-node Kubernetes cluster implemented as a single Docker image that runs as a container.
```
scholzj: My understanding was that Kid could do multinode clusters as well. Is it not the case?

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+It was primarily designed for testing Kubernetes itself, but may be used for local development or CI.
+
+When using a local install of Minikube or Minishift, the Kubernetes cluster is started inside a virtual machine, running a Linux kernel and a Docker daemon, consuming extra CPU and RAM.
+
+Kind, on the other hand, requires no additional VM - it simply runs as a linux container with a set of processes using the same Linux kernel used by your Docker daemon.
```
scholzj: I wonder how much this really applies outside of Linux ... VM is still a VM and I do not think this is really more efficient on MacOS or Windows. Should we maybe add something like _especially on Linux_?

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+You can also follow the operator's log:
+```shell
+kubectl logs `kubectl get pod -n kafka | grep strimzi-cluster-operator | awk '{print $1}'` -n kafka -f
+```
+
+# Provision the Apache Kafka cluster
+
+After that we feed Strimzi with a simple **Custom Resource**, which will then give you a small persistent Apache Kafka Cluster with one node for each, Apache Zookeeper and Apache Kafka:
```
scholzj: I think the _we feed Strimzi_ is a bit weird on a website. I wonder if we can have some better wording here.

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+If there is a mismatch between namespaces, then the Strimzi Cluster Operator will not have the necessary permissions to perform its operations. 
+
+Follow the deployment of strimzi kafka operator:
+```shell
+kubectl get pod -n kafka --watch
+```
+
+You can also follow the operator's log:
+```shell
+kubectl logs `kubectl get pod -n kafka | grep strimzi-cluster-operator | awk '{print $1}'` -n kafka -f
```
scholzj: I wonder if this would be enough without th eneed to have `awk` installed. I'm not sure how muh it is standard utility even on Windows Linux subsystem etc.

```suggestion
kubectl logs `kubectl get pod -n kafka -o name | grep strimzi-cluster-operator` -n kafka -f
```

quickstarts/kind/index.md
mstruk: You're right. I'll change the sentence.

quickstarts/kind/index.md
mstruk: In my experience on MacOS it still makes a difference.
Sure, we can add 'especially on Linux'.

quickstarts/kind/index.md
mstruk: I just copied from Minikube quickstart, and the text is also the same in OKD. What if instead of:

'After that we feed Strimzi with a simple Custom Resource, which will then give you a small persistent Apache Kafka Cluster with one node for each, Apache Zookeeper and Apache Kafka'

We say:

'Then we create a new Kafka custom resource, which will give us a small persistent Apache Kafka Cluster with one node for each - Apache Zookeeper and Apache Kafka:'

quickstarts/kind/index.md
mstruk: That would work as well.

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+layout: default
+---
+
+Kubernetes Kind is a single-node Kubernetes cluster implemented as a single Docker image that runs as a container.
+It was primarily designed for testing Kubernetes itself, but may be used for local development or CI.
+
+When using a local install of Minikube or Minishift, the Kubernetes cluster is started inside a virtual machine, running a Linux kernel and a Docker daemon, consuming extra CPU and RAM.
+
+Kind, on the other hand, requires no additional VM - it simply runs as a linux container with a set of processes using the same Linux kernel used by your Docker daemon.
+For this reason it is faster to start, and consumes less CPU, and RAM than the alternatives.
```
tombentley:
```suggestion
For this reason it is faster to start, and consumes less CPU and RAM than the alternatives.
```

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+# Installing the dependencies
+
+This quickstart assumes that you have the latest version of the `kind` binary, which you can get [here](https://github.com/kubernetes-sigs/kind/releases).
+
+Kind requires a running Docker Daemon. There are different Docker options depending on your host platform.
+You can follow the instructions [here](https://docs.docker.com/get-docker/).
+
+You'll also need `kubectl` binary, which you can get by following the instructions [here](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
```
tombentley:
```suggestion
You'll also need the `kubectl` binary, which you can get by following the instructions [here](https://kubernetes.io/docs/tasks/tools/install-kubectl/).
```

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+kind version
+
+# Validate kubectl
+kubectl version
+```
+
+# Configuring the Docker daemon
+
+If your Docker Daemon runs as a VM you'll most likely need to configure how much memory the VM should have, how many CPUs, how much disk space, and swap size.
+Make sure to assign at least 2 CPUs, and preferably 4 Gb or more of RAM. Consult Docker documentation for you platform.
```
tombentley:
```suggestion
Make sure to assign at least 2 CPUs, and preferably 4 Gb or more of RAM. Consult the Docker documentation for you platform.
```

.. but consult the documentation about what, exactly?

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+Next we apply the Strimzi install files, including `ClusterRoles`, `ClusterRoleBindings` and some **Custom Resource Definitions** (`CRDs`). The CRDs define the schemas used for declarative management of the Kafka cluster, Kafka topics and users.
+
+```shell
+curl -L https://github.com/strimzi/strimzi-kafka-operator/releases/download/{{site.data.releases.operator[0].version}}/strimzi-cluster-operator-{{site.data.releases.operator[0].version}}.yaml \
+  | sed 's/namespace: .*/namespace: kafka/' \
+  | kubectl apply -f - -n kafka 
+```
+
+Note how we set all the `namespace` references in downloaded .yaml file to `kafka`. By default they are set to `myproject`.
```
tombentley:
```suggestion
This will be familiar if you've installed Strimzi on things like minikube before.
Note how we set all the `namespace` references in downloaded .yaml file to `kafka`. By default they are set to `myproject`.
```

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+But we want them all to be `kafka` because we decided to install the operator into `kafka` namespace, which we achieve by specifying `-n kafka` when running `kubectl apply` ensuring that all the definitions and the configurations are installed in `kafka` namespace rather than the `default` namespace.
```
tombentley:
```suggestion
But we want them all to be `kafka` because we decided to install the operator into the `kafka` namespace, which we achieve by specifying `-n kafka` when running `kubectl apply` ensuring that all the definitions and the configurations are installed in `kafka` namespace rather than the `default` namespace.
```

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+
+If there is a mismatch between namespaces, then the Strimzi Cluster Operator will not have the necessary permissions to perform its operations.
```
tombentley: Maybe we should mention the error the user would see, since we regular have to help people on this point on Slack.

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+  | sed 's/namespace: .*/namespace: kafka/' \
+  | kubectl apply -f - -n kafka 
+```
+
+Note how we set all the `namespace` references in downloaded .yaml file to `kafka`. By default they are set to `myproject`.
+But we want them all to be `kafka` because we decided to install the operator into `kafka` namespace, which we achieve by specifying `-n kafka` when running `kubectl apply` ensuring that all the definitions and the configurations are installed in `kafka` namespace rather than the `default` namespace.
+
+If there is a mismatch between namespaces, then the Strimzi Cluster Operator will not have the necessary permissions to perform its operations. 
+
+Follow the deployment of strimzi kafka operator:
```
tombentley:
```suggestion
Follow the deployment of the Strimzi Kafka operator:
```

quickstarts/kind/index.md
tombentley: SGTM

quickstarts/kind/index.md
```diff
@@ -0,0 +1,112 @@
...
+kubectl logs `kubectl get pod -n kafka | grep strimzi-cluster-operator | awk '{print $1}'` -n kafka -f
+```
+
+# Provision the Apache Kafka cluster
+
+After that we feed Strimzi with a simple **Custom Resource**, which will then give you a small persistent Apache Kafka Cluster with one node for each, Apache Zookeeper and Apache Kafka:
+
+```shell
+# Apply the `Kafka` Cluster CR file
+kubectl apply -f https://raw.githubusercontent.com/strimzi/strimzi-kafka-operator/{{site.data.releases.operator[0].version}}/examples/kafka/kafka-persistent-single.yaml -n kafka
```
tombentley: can't we use `latest` to avoid the parameter?

quickstarts/kind/index.md
scholzj: Yeah, I think that would be better.

tombentley:
You'll need to signoff the commits.

quickstarts/kind/index.md
mstruk: This targets an existing branch or tag. We have `master` branch but that's a SNAPSHOT version, not a released version. We could create a `latest` branch and always keep it equal to latest released version. In that case we need to make it part of the release procedure to also hard reset the 'latest' branch.

quickstarts/kind/index.md
mstruk: What if we create Troubleshooting.md document where we collect all sorts of troubles, and add a link here to point to the exact problem we have in mind?

mstruk: There is an [existing document](https://github.com/strimzi/strimzi-kafka-operator/blob/master/documentation/modules/appendix_faq.adoc) that seems a fitting place for this. Or not? That document is rather hard to read ...

quickstarts/kind/index.md
```diff
@@ -0,0 +1,113 @@
...
+---
+
+Kubernetes Kind is a Kubernetes cluster implemented as a single Docker image that runs as a container.
+It was primarily designed for testing Kubernetes itself, but may be used for local development or CI.
+
+When using a local install of Minikube or Minishift, the Kubernetes cluster is started inside a virtual machine, running a Linux kernel and a Docker daemon, consuming extra CPU and RAM.
+
+Kind, on the other hand, requires no additional VM - it simply runs as a linux container with a set of processes using the same Linux kernel used by your Docker daemon.
+For this reason it is faster to start, and consumes less CPU and RAM than the alternatives, which is especially noticeable when running a Docker daemon natively on a Linux host.
+
```
scholzj:
```suggestion

_Note: Kubernetes Kind does currently not support node ports or load balancers. You will not be able to easily access your Kafka cluster from outside of the Kubernetes environment. If you need access from outside, we recommend to use Minikube instead._

```

PR #134, opened 2020-04-08
Fixed link to what is an operator

ppatierno:
This PR fixes the link to a "what is an operator" article.

### DIFF
--- a/_includes/homepage/homepage-hero-band.html
+++ b/_includes/homepage/homepage-hero-band.html
@@ -35,7 +35,7 @@
         <p>
           <ul>
             <li>For example kubectl get kafka</li>
-            <li>Operator Based (<a href="https://kubernetes.io/docs/concepts/extend-kubernetes/operator/">What is an operator</a>)</li>
+            <li>Operator Based (<a href="https://kubernetes.io/docs/concepts/extend-kubernetes/operator/">What is an operator?</a>)</li>
             <li>Manage Kafka using gitops</li>
           </ul>
         </p>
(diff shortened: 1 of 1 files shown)


tombentley:
Why not https://kubernetes.io/docs/concepts/extend-kubernetes/operator/, which is the top-ranking google result for "what is a kubernetes operator"?

ppatierno:
@tombentley yeah agree, much better even because coming from Kubernetes. Just updated.

_includes/homepage/homepage-hero-band.html
```diff
@@ -35,7 +35,7 @@ <h4>Kubernetes-Native Experience</h4>
         <p>
           <ul>
             <li>For example kubectl get kafka</li>
-            <li>Operator Based (<a href="{{site.baseurl}}/docs/overview/latest/">What is an operator</a> and why should I use them?)</li>
+            <li>Operator Based (<a href="https://kubernetes.io/docs/concepts/extend-kubernetes/operator/">What is an operator</a> and why should I use them?)</li>
```
scholzj, line 38: I rasied it already 10 times, but I will do it again because I'm persistent. I think we should remove the `and why should I use them` part. It makes this point to be on two lines which looks bad to me on the website and I do not think it really adds any value.

ppatierno: and I forgot it 11 times ;-) Just fixed it.

scholzj: Should we keep the `?` ... `What is an operator?`?

ppatierno: Done

PR #146, opened 2020-04-23
Update copyright footer

lucperkins:
Doesn't look super great. Maybe someone who's more acquainted with the site's CSS can follow up and make it look better.

### DIFF
--- a/_includes/project-footer.html
+++ b/_includes/project-footer.html
@@ -17,8 +17,11 @@
     </div>
     <div class="width-3-12 width-12-12-m justify-self-center align-self-center">
       <p>
-        Strimzi is licensed under the<br/>
-        <a href="https://strimzi.io/LICENSE">Apache Licence, Version 2.0</a>
+        &copy; Strimzi Authors {{ 'now' | date: "%Y" }} | Documentation distributed under <a href="https://creativecommons.org/licenses/by/4.0">CC-BY-4.0</a>
+      </p>
+
+      <p>
+        &copy; {{ 'now' | date: "%Y" }} The Linux Foundation. All rights reserved. The Linux Foundation has registered trademarks and uses trademarks. For a list of trademarks of The Linux Foundation, please see our <a href="https://www.linuxfoundation.org/trademark-usage/">Trademark Usage</a> page.
       </p>
     </div>
   </div>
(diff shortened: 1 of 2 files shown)


tombentley:
Thanks @lucperkins. I think we need a link to the LF privacy policy too (https://www.linuxfoundation.org/privacy/), would you be able to add that?

_includes/project-footer.html
```diff
@@ -17,8 +17,11 @@
     </div>
     <div class="width-3-12 width-12-12-m justify-self-center align-self-center">
       <p>
-        Strimzi is licensed under the<br/>
-        <a href="https://strimzi.io/LICENSE">Apache Licence, Version 2.0</a>
+        &copy; Strimzi Authors {{ 'now' | date: "%Y" }} | Documentation distributed under <a href="https://creativecommons.org/licenses/by/4.0">CC-BY-4.0</a>
```
scholzj, line 20: This now gives the website two copyrights - one for Strimzi Authors and another one for CNCF. Does that actually work? I would assume the page should have only one copyright, or? I can see that for example kubernetes.io has the same, but it still looks strange.

_includes/project-footer.html
```diff
@@ -17,8 +17,11 @@
...
+      </p>
+
+      <p>
+        &copy; {{ 'now' | date: "%Y" }} The Linux Foundation. All rights reserved. The Linux Foundation has registered trademarks and uses trademarks. For a list of trademarks of The Linux Foundation, please see our <a href="https://www.linuxfoundation.org/trademark-usage/">Trademark Usage</a> page.
```
scholzj, line 24: When I look at the preview (https://deploy-preview-146--strimzi.netlify.app/), this looks quite weird because this text is very long. I think it might be better to move the CNCF copyright and trademark usage to a new line spnning the whole site. WDYT?

lucperkins: As expressed in the PR description, I was hoping that one of the maintainers who's familiar with the CSS framework could follow up and bring it in line with the site's aesthetic standards

scholzj: Ok. Did you allowed maintainers to push into this PR?

lucperkins: Certainly! No problem at all.

lucperkins: @scholzj Also please feel free to close this and create your own PR if that's easier.

lucperkins:
LGTM

PR #148, opened 2020-04-24
Add namespace changing function

scholzj:
This PR adds a new function to our website which can be used to change the namespace in the RoleBinding / ClusterRoleBinding subjects. It can be called to modify the installation files instead of using `sed`. So something like this:

```
curl -L https://github.com/strimzi/strimzi-kafka-operator/releases/download/0.17.0/strimzi-cluster-operator-0.17.0.yaml \
  | sed 's/namespace: .*/namespace: kafka/' \
  | kubectl apply -f - -n kafka 
```

can be replaced by this:

```
kubectl apply -f https://strimzi.io/install/latest?namespace=kafka -n kafka
```

There are some gotchas to think about ...

1) The fucntion really does simple search and replace in the installation files. So it sets the desired namespace only when we already have the namespace. As a result, the namespace needs to be specified twice: once in the URL and once in kubectl. To fix this, we would either need to add namespace to all installation files or create more sophisticated function which would parse the YAML files and replace/add the namespace that way.

2) The namespace is currently passed as a URL paramater. But `?` needs to be normally escaped in terminal. In my case it seems to work fine. My terminal escapes it on the fly when copy pasting. Not sure it works that way always. Sot maybe we should consider something simple? Such as maybe `https://strimzi.io/install/latest/<namespace>`? That could be done, but it might be a bit more confusing for the user.

WDYT? How should we proceed with this?


tombentley:
Thanks for looking into this idea @scholzj.

1. You're talking about `metadata.namespace` in particular, I assume. That's a good point. IDK whether just adding the namespace to every resource might cause problems with the helm install or the CSVs. If not then I suppose that might be the simplest way to do it. But I feel like parsing the YAML would be less likely to result in unintended consequences. Surely there's some node module for YAML which would make this fairly easy?

2. I think `https://strimzi.io/install/latest/<namespace>` would indeed be confusing, and potentially constrains future website reorgs because of the need to maintain compatibility. That concern could be mitigated by using a subdomain `https://install.strimzi.io/latest/<namespace>` (or `https://install.strimzi.io/latest/namespace/<namespace>`) but that would need further hoop jumping to set up. I guess I'd be willing to live with the need for quoting.

scholzj:
* I have no idea how to add the subdomain. Not sure Netlify can handle this. So I leave that up to someone else.
* I updated the code to be able to do just `kubectl apply -f https://strimzi.io/install/latest?namespace=kafka`. That is probably all I can do.

quickstarts/kind.md
```diff
@@ -55,9 +55,7 @@ kubectl create namespace kafka
 Next we apply the Strimzi install files, including `ClusterRoles`, `ClusterRoleBindings` and some **Custom Resource Definitions** (`CRDs`). The CRDs define the schemas used for declarative management of the Kafka cluster, Kafka topics and users.
 
 ```shell
-curl -L https://github.com/strimzi/strimzi-kafka-operator/releases/download/{{site.data.releases.operator[0].version}}/strimzi-cluster-operator-{{site.data.releases.operator[0].version}}.yaml \
-  | sed 's/namespace: .*/namespace: kafka/' \
-  | kubectl apply -f - -n kafka 
+kubectl apply -f https://strimzi.io/install/latest?namespace=kafka -n kafka
```
tombentley, line 58: Should we quote those?

scholzj: Not sure. Does it happen automatically for you when you copy paste them into terminal? It does for me, but not sure how much this is common future or some MacOS trick.

tombentley: It doesn't get quoted in my (fariyl bog standard) Fedora setup. Presumably there'll be some terminal emulators which do support it. Thinking more, I guess the chances of the shell expanding the `?` are actually pretty close to zero, since it's so unlikely there'll be matching filenames. So I suppose we can leave it like this and fix it if anyone ever complains.

scholzj: So, just to make sure the stuff works weel for you ... would this work fine for you?

```
kubectl apply -f https://strimzi.io/install/latest\?namespace\=kafka
```

tombentley: Yes. Single quotes works too. And just plain`?` works if the shell can't expand the wildcard.

PR #151, opened 2020-04-29
update logs command to use deployment

hguerrero:
use deploymet approach instead of doing grep.


scholzj:
Thanks for the PR @hguerrero

PR #174, opened 2020-07-25
Add 0.19.0-rc1 Helm Charts to the index

scholzj:
As described in strimzi/strimzi-kafka-operator#3380, Helm should be able to deal with RC versions. When `--devel` is specified, it will use RCs and otherwise it will use stable only. We should give it a try as that might make easier for anyone to test the RC versions using Helm. This PR adds the Helm 2 and 3 charts to the index to see how well it works.

### DIFF
--- a/charts/index.yaml
+++ b/charts/index.yaml
@@ -2,6 +2,60 @@ apiVersion: v1
 entries:
   strimzi-kafka-operator:
   - apiVersion: v1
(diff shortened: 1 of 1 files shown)


scholzj:
Closing this PR since we will have the full 0.19.0 release before this is approved. We should do this next time in 0.20.0.

PR #185, opened 2020-09-05
Removed Strimzi survey band

ppatierno:
This PR removes the Strimzi survey band. I preferred to not delete the related HTML file but to comment the content in order to have it ready as a placeholder when we have to do another announcement; I preferred this compared to comment the "include" in the home as well.

### DIFF
--- a/_includes/homepage/homepage-announcement-band.html
+++ b/_includes/homepage/homepage-announcement-band.html
@@ -1,4 +1,3 @@
-<!--
 <div class="component-slim announcement_orange">
   <div class="grid-wrapper">
     <div class="width-12-12">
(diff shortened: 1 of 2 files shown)


scholzj (commented):
Can you maybe comment out the place where this file is included instead? It will make the website smaller ;-)

ppatierno:
I didn't want to lose the content for using it when needed. Are you really concerned about few bytes?

scholzj:
I'm not suggesting to delete the file - we should keep it to reuse it next time. But the file seems to contain only the banner. So it has to be somewhere included into the main page. And if you comment out the include command, we will have this file there , but it will not load into the website.

scholzj (approved):

PR #209, opened 2020-11-19
Add new section to Download page with supported versions of the different components

scholzj:
Following the idea from the _Meet the Maintainers_ session, this PR drafts how a new section on the downloads page might look like (content is just dummy right now). Please have a look and let me know what do you think.

I guess other alternative would be to have a separate page with the versions table and only link it from the downloads page.

### DIFF
--- a/_data/releases.yaml
+++ b/_data/releases.yaml
@@ -1,6 +1,10 @@
 operator:
   - version: 0.20.0
     defaultKafkaVersion: 2.6.0
+    bridgeVersion: 0.19.0
+    oAuthVersion: 0.6.1
+    kafkaVersions: 2.5.0, 2.5.1, 2.6.0
+    kubernetesVersions: 1.11+
     overview_book: true
     quickstart_book: true
     using_book: true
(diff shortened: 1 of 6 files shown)


tombentley (approved):
This looks great. One minor comment which you can take or leave as you see fit: The order of the Kubernetes versions in the versions matrix was the opposite of what I would expect.

ppatierno (approved):
LGTM. @tombentley why you would expect the opposite order? Having the last one at the top is quite common, or?

tombentley:
@ppatierno I just found it surprising. I agree that it's not wrong.

scholzj:
@tombentley What do you mean with _The order of the Kubernetes versions_? It is just 1.11+ or 1.16+. Did you meant to have the whole table in different order with the oldest operator versions on top or newest at the bottom?

tombentley:
Sorry, I meant Kafka versions.

ppatierno:
Well, reading from left to right I would expect the more recent Kafka version to be the first.
The same for Kubernetes reading from top to bottom, having more recent at the top.

scholzj:
Ahh, ok ... so `2.5.0, 2.5.1, 2.6.0` instead of `2.6.0, 2.5.1, 2.5.0`. Yeah, that probably makes sense.

ppatierno:
Tbh I am for the current proposal with newer in the left.

scholzj:
@ppatierno @tombentley I updated it to the proper styles handling and generating the table from the release YAML. I added the info back to 0.12. I think that should be enough. Please have another look and let me know what do you think.

ppatierno:
+1

tombentley:
LGTM.

Frawless (approved):

PR #212, opened 2020-12-21
[POST] consumer configuration tuning

PaulRMellor:

New post to describe consumer optimization options


_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+
+### Message ordering guarantees
+
+Kafka *only* provides ordering guarantees for messages in a single partition. If you want a strict ordering of messages from one topic, the only option is to use one partition per consumer. The consumer can then observe messages in the same order that they were committed to the broker.
+
+Things become less precise when a consumer is consuming messages from multiple partitions. The order of messages fetched from different partitions does not necessarily reflect the order in which they were sent.
+
+### Scaling with consumer groups
+
+You'll need a group id if you intend to use consumer groups. Consumer groups are used so commonly, that they might be considered part of a basic consumer configuration. Consumer groups share a data stream generated by multiple producers from a given topic. Consumers are grouped using a `group.id`, allowing messages to be spread across the members that share the same id.
```
scholzj, line 72: > Consumer groups share a data stream generated by multiple producers from a given topic.

Would _Consumer group **members** share a data stream generated by multiple producers from a given topic._? Because I do not think I would say that different consumer groups share the data ... it is always the consumers which are part of a single group who shares the data.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+
+```properties
+# ...
+group.id=my-group-id
+# ...
+```
+
+Consumer groups are very useful for scaling your consumers according to demand. Consumers within a group do not read data from the same partition, but can receive data exclusively from one or more partitions. As mentioned previously, if you want to guarantee message ordering, you would use one consumer with a single partition.
```
scholzj, line 80: > but can receive data exclusively from one or more partitions

Technically it is _zero or more partitions_ since a consumer can be assigned 0 partitions as well. I think we should also stress out here that one partition is assigned only to one member of the consumer group, never more.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+
+There are a couple of things worth mentioning when thinking about how many consumers to include in a group. Adding more consumers than partitions will not increase throughput, but the additional consumers will increase the request load for the broker that serves the consumer group. On the other hand, there is a potential upside in terms of reliability, because there will be consumers on standby in the event of failure.
```
scholzj, line 82: > but the additional consumers will increase the request load for the broker that serves the consumer group

@tombentley Is there any load significant enough to be worth mentioning? I guess all they do when no partitions is assigned is just heartbeats for the consumer group, or?

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+: The `commitAsync` API does not wait for the broker to respond to a commit request. The `commitAsync` API has lower latency than the `commitSync` API, but risks creating duplicates when rebalancing.
+
+A common approach is to capitalize on the benefits of using both APIs, so the lower latency `commitAsync` API is used by default, but the `commitSync` API takes over before shutting the consumer down or rebalancing to safeguard the final commit.
+
+## Configuring a reliable data pipeline
+
+To guarantee the reliability of message delivery on the _producer side_, you might configure your producers to use idempotence and transactional ids.
+
+Idempotence
+: Idempotence properties used for idempotence guarantee the order of message delivery from producers for exactly once writes to a single partition.
```
tombentley: Idempotence isn't directly about the order of messages, but avoiding duplicates due to retried produce requests.
```suggestion
: Idempotence properties used to guarantee the non-duplication and ordering of message delivery from producers for exactly once writes to a single partition.
```

scholzj (commented):
Looks great. Just some minor comments about some part which I think might be a bit confusing for a newbie users.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+isolation.level=read_committed
+# ...
+```
+
+## Recovering from failure within a consumer group
+
+You can define how often checks are made on the health of consumers within a consumer group. If consumers fail within a consumer group, a rebalance is triggered and partition ownership is reassigned to the members of the group. You want to get the timings of your checks just right so that the consumer group can recover quickly, but unnecessary rebalances are not triggered. And you use two properties to do it: `session.timeout.ms` and `heartbeat.interval.ms`.
+
+`session.timeout.ms`
+: Specifies the maximum amount of time in milliseconds a consumer within a consumer group can be out of contact with a broker before being considered inactive and a _rebalancing_ is triggered between the active consumers in the group. Setting the `session.timeout.ms` property lower means failing consumers are detected earlier, and rebalancing can take place quicker. However, you don't want to set the timeout so low that the broker fails to receive an _all-clear_ heartbeat in time and triggers an unnecessary rebalance.
```
tombentley: What is an _all-clear_ heartbeat. I've never seen this term before.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Managing offset policy
+
+How should a consumer behave when no offsets have been committed? Or when a committed offset is no longer valid or deleted? If you set the `auto.offset.reset` property correctly, it should behave impeccably in both events.
+
+Let's first consider what should happen when no offsets have been committed. Suppose a new consumer application connects with a broker and presents a new consumer group id for the first time. Offsets determine up to which message in a partition a consumer has read from. Consumer offset information lives in an internal Kafka topic called `__consumer_offsets`. The `__consumer_offsets` topic does not yet contain any offset information for this new application. So where does the consumer start from? With the `auto.offset.reset` property set as `latest`, which is the default, the consumer will start processing only new messages. By processing only new messages, some messages will be missed. To avoid data loss, you can set the `auto.offset.reset` property to `earliest` and also process existing messages from the start of the log.
```
tombentley, line 227:
```suggestion
Let's first consider what should happen when no offsets have been committed. Suppose a new consumer application connects with a broker and presents a new consumer group id for the first time. Offsets determine up to which message in a partition a consumer has read from. Consumer offset information lives in an internal Kafka topic called `__consumer_offsets`. The `__consumer_offsets` topic does not yet contain any offset information for this new application. So where does the consumer start from? With the `auto.offset.reset` property set as `latest`, which is the default, the consumer will start processing only new messages. By processing only new messages, any existing messages will be missed. Alternatively, you can set the `auto.offset.reset` property to `earliest` and also process existing messages from the start of the log.
```

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+
+And what happens when offsets are no longer valid? If a consumer group or standalone consumer is inactive and commits no offsets during the _offsets retention period_ (`offsets.retention.minutes`) configured for a broker, previously committed offsets are deleted from `__consumer_offsets`. Again, you can use the `earliest` option in this situation so that the consumer returns to the start of a partition to avoid data loss if offsets were not committed.
+
+If the amount of data returned in a single fetch request is large, a timeout might occur before the consumer has processed it. In which case, you can lower `max.partition.fetch.bytes` or increase `session.timeout.ms` as part of you offset policy.
```
tombentley, line 231: I think it would be good to reword this to make it clearer that it's the frequency with which the application calls poll() that determines whether there's a timeout.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+
+```properties
+# ...
+group.instance.id=UNIQUE-ID
+# ...
+```
+
+The consumer group coordinator assigns the consumer instance a new member id, but as a static member it continues with the same instance id, and receives the same assignment of topic partitions is made.
+
+You can also calibrate the `max.poll.interval.ms` and `max.poll.interval.ms` polling configuration properties to reduce the number of rebalances.
```
tombentley, line 262: Changing these only really helps when the cause of the rebalances is due to the poll interval. IOW, it's only indicated when the logs demontrated that this is the problem.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+# ...
+group.instance.id=UNIQUE-ID
+max.poll.interval.ms=300000
+max.poll.records=500
+# ...
+```
+
+### Changing your consumer habits
+
+So that was a brief run through some of the most frequently used configuration options. Any strategy you choose to adopt will always have to take into account the other _moving parts_ -- the intended functionality of your consumer client application, the broker architecture, and the producer configuration. For example, if you are not using transactional IDs with your producers, then you probably won't see a need for using the `isolation.level` property. But it is unlikely that there isn't anything you can do to make your consumers run better. And the consumer configuration options are available exactly for that reason.
```
tombentley, line 280:
```suggestion
So that was a brief run through some of the most frequently used configuration options. Any strategy you choose to adopt will always have to take into account the other _moving parts_ -- the intended functionality of your consumer client application, the broker architecture, and the producer configuration. For example, if you are not using transactional producers, then there's no point in setting the `isolation.level` property. But it is unlikely that there's  nothing you can do to make your consumers run better. And the consumer configuration options are available exactly for that reason.
```

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+> Be careful with memory usage. The maximum amount of memory a client can consume is calculated approximately as:
+>
+> `_NUMBER-OF-BROKERS_ * fetch.max.bytes and _NUMBER-OF-PARTITIONS_ * max.partition.fetch.bytes`
+
+
+## Avoiding data loss or duplication when committing offsets
+
+For applications that require durable message delivery, you can increase the level of control over consumers when committing offsets to minimize the risk of data being lost or duplicated.
+
+Kafka's _auto-commit mechanism_ is pretty convenient (and sometimes suitable). When enabled, consumers commit the offsets of messages automatically. But convenience, as always, has a price. By allowing your consumer to commit offsets, you are introducing a risk of data loss and duplication. Why? Quite easily as it happens. You see, the auto-commit can _only_ avoid data loss if all messages are processed before the next poll of the broker or the consumer closes. Using the auto-commit, the broker is polled at 5000ms intervals. In that time, the inherent constraints of auto-committing might lead to:
```
ppatierno, line 141: `Using the auto-commit, the broker is polled at 5000ms intervals.` ... that's wrong.
The polling time is defined by the application (the consumer has often a loop calling the poll). I guess that you are referring the default interval of committing offsets using auto-commit. Let's avoid to use the "poll" word that means something more specific in Kafka land.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+# ...
+enable.auto.commit=false
+auto.commit.interval.ms=1000
+# ...
+```
+
+A more robust course of action, away from consumer configuration, is to set up your consumer client application to commit offsets after _all_ processing has been performed and messages have been consumed. To do this, you can introduce calls to the Kafka `commitSync` and `commitAsync` APIs, which commit specified offsets for topics and partitions to Kafka.
+
+* **`commitSync`**
+The `commitSync` API will not poll for new messages until the last offset in a message batch is committed. This can affect throughput and latency, but you can configure the consumer to commit less frequently.
```
ppatierno, line 158: I don't think that this API is related to poll or it avoids you to poll for new messages while committing.
I mean, if I write an application where in two different threads I do commitSync and poll I guess I can do that. Of course, it's just an example, something I would not do but just to explain that the sentence sounds misleading to me. Maybe it's still related to the fact that we should pay more attention when the "poll" verb is used because it has a specific meaning in Kafka. @tombentley wdyt?

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+isolation.level=read_committed
+# ...
+```
+
+## Recovering from failure within a consumer group
+
+You can define how often checks are made on the health of consumers within a consumer group. If consumers fail within a consumer group, a rebalance is triggered and partition ownership is reassigned to the members of the group. You want to get the timings of your checks just right so that the consumer group can recover quickly, but unnecessary rebalances are not triggered. And you use two properties to do it: `session.timeout.ms` and `heartbeat.interval.ms`.
+
+* **`session.timeout.ms`**
+Specifies the maximum amount of time in milliseconds a consumer within a consumer group can be out of contact with a broker before being considered inactive and a _rebalancing_ is triggered between the active consumers in the group. Setting the ** * * `session.timeout.ms` property lower means failing consumers are detected earlier, and rebalancing can take place quicker. However, you don't want to set the timeout so low that the broker fails to receive an _all-clear_ heartbeat in time and triggers an unnecessary rebalance.
```
ppatierno, line 209: "** * * `session.timeout.ms`" too many asterisks?

ppatierno (commented):
I left some comments. As usual it would be better having one sentence per line to be more precise with comment.s

_posts/2021-01-08-consumer-tuning.md
tombentley: Yeah, this sentence is confusing. `commitSync` doesn't call `poll()`, but rather commits the offsets of all the messages which were returned by the last `poll()`. " you can configure the consumer to commit less frequently" is also a bit ambiguous. I assume the context here is a consumer application where we're doing a "manual commit", in which case it's not really configuration, but rather how that application is coded. I.e. is it committing after every poll, or less frequently. And how many messages are returned by each poll? 

We also don't talk about an important alternative to committing offsets in Kafka when interacting with some other system which supports ACID transactions. In that case you can get better processing guarantees by storing the offsets in that system. I'm not saying we should talk about this in detail, because the whole topic can mushroom into its own blog post, merely highlighting that it might be worth mentioning in a sentence.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+Sets a maximum threshold for time-based batching.
+
+* **`fetch.min.bytes`**
+Sets a minimum threshold for size-based batching.
+
+When the client application polls for data, both these properties govern the amount of data fetched by the consumer from the broker. You can adjust the properties higher so that there are fewer requests, and messages are delivered in bigger batches. You might want to do this if the amount of data being produced is low. Reducing the number of requests from the consumer also lowers the overhead on CPU utilization for the consumer and broker.
+
+Throughput is improved by increasing the amount of data fetched in a batch, but with some cost to latency. Consequently, adjusting these properties lower has the effect of lowering end-to-end latency. We'll look a bit more at targeting latency by increasing batch sizes in the next section.
+
+You can use one or both or these properties. If you use both, Kafka will respond to a fetch request when the first of either threshold is reached.
```
laidan6000, line 104:
```suggestion
You can use one or both of these properties. If you use both, Kafka will respond to a fetch request when the first of either threshold is reached.
```

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+```properties
+# ...
+fetch.max.bytes=52428800
+max.partition.fetch.bytes=1048576
+# ...
+```
+
+> Be careful with memory usage. The maximum amount of memory a client can consume is calculated approximately as:
+>
+> `_NUMBER-OF-BROKERS_ * fetch.max.bytes and _NUMBER-OF-PARTITIONS_ * max.partition.fetch.bytes`
```
laidan6000, line 134: Correct markup for italics?

laidan6000 (approved):
This is great content!

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,280 @@
...
+
+As with producers, you will want to achieve a balance between throughput and latency that meets your needs. You will also want to ensure your configuration helps reduce delays caused by unnecessary rebalances of consumer groups. Note, however, that you should avoid using any properties that cause conflict with the properties or guarantees provided by your application.
+
+>If you want to read more about what each property does, see Kafka's [consumer configs](https://kafka.apache.org/documentation/#consumerconfigs).
+
+### Message ordering guarantees
+
+Kafka *only* provides ordering guarantees for messages in a single partition. If you want a strict ordering of messages from one topic, the only option is to use one partition per consumer. The consumer can then observe messages in the same order that they were committed to the broker.
+
+Things become less precise when a consumer is consuming messages from multiple partitions. The order of messages fetched from different partitions does not necessarily reflect the order in which they were sent.
```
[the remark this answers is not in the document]
PaulRMellor: I changed the first paragraph to _per topic_.  For the second paragraph, I've updated as follows:

> Potentially, things become less precise when a consumer is consuming messages from multiple partitions. Although the ordering for each partition is kept, the order of messages fetched from _all_ partitions is not guaranteed, as it does not necessarily reflect the order in which they were sent. In practice, however, retaining the order of each partition when consuming from multiple partitions is sufficient for most well-designed consumer client applications.

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: >  Consumer **group** members share a data stream generated by multiple producers from a given topic.

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Updates:

> One partition is assigned to exactly one member of a consumer group. 

> There are a couple of things worth mentioning when thinking about how many consumers to include in a group. Adding more consumers than partitions will not increase throughput, but additional consumers will increase the request load for the broker that serves the consumer group. And any excess consumers will be partition-free and idle. On the other hand, there is a potential upside in terms of reliability, because those idle consumers will be on standby in the event of failure.

ppatierno: `but additional consumers will increase the request load for the broker that serves the consumer group` I am still not convinced mentioning this, because it should be just an heartbeat.

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Updated as directed

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+
+```properties
+# ...
+group.id=my-group-id
+# ...
+```
+
+Consumer groups are very useful for scaling your consumers according to demand. Consumers within a group do not read data from the same partition, but can receive data exclusively from one or more partitions. As mentioned previously, if you want to guarantee message ordering, you would use one consumer with a single partition.
+
+There are a couple of things worth mentioning when thinking about how many consumers to include in a group. Adding more consumers than partitions will not increase throughput, but the additional consumers will increase the request load for the broker that serves the consumer group. On the other hand, there is a potential upside in terms of reliability, because there will be consumers on standby in the event of failure.
```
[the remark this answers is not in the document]
PaulRMellor: Updated to:

> There are a couple of things worth mentioning when thinking about how many consumers to include in a group. Adding more consumers than partitions will not increase throughput. Excess consumers will be partition-free and idle. On the other hand, there is a potential upside in terms of reliability, because those idle consumers will be on standby in the event of failure.

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+
+### Consumer lag and consumer groups
+
+For applications that rely on processing (near) real-time data, consumer lag is a vitally important metric. Consumer lag indicates the difference in the rate of production and consumption of messages. Specifically, consumer lag for a given consumer group indicates the delay between the last message added to a topic partition and the message last picked up by the consumer of that partition. If consumer lag is a problem, there are definite actions you can take to reduce it. Adding more disk capacity or increasing the retention time for a message to remain in a topic might help. Concerning consumer tuning, you can add more consumers to the consumer group subscribed to a topic. An important point to make here, though, is that consumer configuration tuning in isolation might not be sufficient to meeting your optimization goals.
```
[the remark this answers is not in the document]
PaulRMellor: Thanks for clarifying @tombentley. Updates made :+1:

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,281 @@
...
+
+* _Data loss_ when the consumer has fetched and transformed a number of messages, but the system crashes with processed messages still in the consumer buffer
+* _Data duplication_ on another consumer instance, after rebalancing, when the system crashes whilst processing the messages, but before performing the auto-commit
+
+If this potential situation leaves you slightly concerned, what can you do about it? First of all, you can simply turn off auto-committing by setting `enable.auto.commit` to `false`. You are now in the driving seat. What else? Alternatively, you can use the `auto.commit.interval.ms` property to decrease those worrying intervals between commits.
```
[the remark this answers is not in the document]
PaulRMellor: I've rearranged, so we talk about the _interval_ first and then switching off _auto-commit_. I've underlined the importance of considering auto or manual, before we describe the APIs in more detail. Not sure what else we might want to say.

image

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Removed!

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Changed to:

> If the amount of data returned in a single fetch request is large, depending on the frequency in which the consumer client application polls for new messages, a timeout might occur before the consumer has processed it. In which case, you can lower `max.partition.fetch.bytes` or increase `session.timeout.ms` as part of you offset policy.

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Changed to:

> Another cause of rebalancing might actually be due to an insufficient poll interval configuration, which is then interpreted as a consumer failure. If you discover this situation when examining your consumer logs, you can calibrate the `max.poll.interval.ms` and `max.poll.interval.ms` polling configuration properties to reduce the number of rebalances.

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Changed to:

> Using the auto-commit, the consumer will commit offsets received from the broker at 5000ms intervals.

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Thanks. Trying to to chop the concepts down so much didn't help here. Hopefully, this get's closer to what we want to say.

> * **`commitSync`**
> The `commitSync` API commits the offsets of all messages returned from polling. Generally, you call the API when you are finished processing all the messages in a batch and don't poll for new messages until the last offset in the batch is committed. This approach can can affect throughput and latency, as can the number of messages returned when polling, so you can set up your application to commit less fequently. 

And I've added a note on ACID.

> You can set up your Kafka solution to interact with databases that provide ACID
> (atomicity, consistency, isolation, and durability) reliability
> guarantees to store consumer offsets.

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Bit of formatting required with `pre` tags.

> Be careful with memory usage. The maximum amount of memory a client can consume is calculated approximately as:
>
> <pre><i>NUMBER-OF-BROKERS</i> * fetch.max.bytes and <i>NUMBER-OF-PARTITIONS</i> * max.partition.fetch.bytes</pre>

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,292 @@
...
+You can use one or both of these properties. If you use both, Kafka will respond to a fetch request when the first of either threshold is reached.
+
+```properties
+# ...
+fetch.max.wait.ms=500
+fetch.min.bytes=16384
+# ...
+```
+
+## Lowering latency by increasing the maximum batch sizes
```
scholzj, line 117: This and some of the following headings suddenly use higher heading levels than the previous ones which used `###` ... I do not think it is intentional, or?

PaulRMellor: I've changed them all to _###_

scholzj (approved):

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,292 @@
...
+
+You'll need a group id if you intend to use consumer groups. Consumer groups are used so commonly, that they might be considered part of a basic consumer configuration. Consumer groups are a way of sharing the work of consuming messages from a set of partitions between a number of consumers by dividing the partitions between them. Consumers are grouped using a `group.id`, allowing messages to be spread across the members that share the same id.
+
+```properties
+# ...
+group.id=my-group-id
+# ...
+```
+
+Consumer groups are very useful for scaling your consumers according to demand. Consumers within a group do not read data from the same partition, but can receive data exclusively from zero or more partitions. One partition is assigned to exactly one member of a consumer group. As mentioned previously, if you want to guarantee message ordering, you would use one consumer with a single partition.
```
tombentley, line 80:
```suggestion
Consumer groups are very useful for scaling your consumers according to demand. Consumers within a group do not read data from the same partition, but can receive data exclusively from zero or more partitions. Each partition is assigned to exactly one member of a consumer group. 
```

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,292 @@
...
+
+There are a couple of things worth mentioning when thinking about how many consumers to include in a group. Adding more consumers than partitions will not increase throughput. Excess consumers will be partition-free and idle. On the other hand, there is a potential upside in terms of reliability, because those idle consumers will be on standby in the event of failure.
```
tombentley, line 82:
```suggestion
There are a couple of things worth mentioning when thinking about how many consumers to include in a group. Adding more consumers than partitions will not increase throughput. Excess consumers will be partition-free and idle. This might not be entirely pointless, however, as those an idle consumer is effectively on standby in the event of failure of one of the consumers that does have partitions assigned.
```

tombentley (approved):

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,292 @@
...
+> Be careful with memory usage. The maximum amount of memory a client can consume is calculated approximately as:
+>
+> <pre><i>NUMBER-OF-BROKERS</i> * fetch.max.bytes and <i>NUMBER-OF-PARTITIONS</i> * max.partition.fetch.bytes</pre>
+
+
+## Avoiding data loss or duplication when committing offsets
+
+For applications that require durable message delivery, you can increase the level of control over consumers when committing offsets to minimize the risk of data being lost or duplicated.
+
+Kafka's _auto-commit mechanism_ is pretty convenient (and sometimes suitable). When enabled, consumers commit the offsets of messages automatically. But convenience, as always, has a price. By allowing your consumer to commit offsets, you are introducing a risk of data loss and duplication. Why? Quite easily as it happens. You see, the auto-commit can _only_ avoid data loss if all messages are processed before the next poll of the broker or the consumer closes. Using the auto-commit, the consumer will commit offsets received from the broker at 5000ms intervals. In that time, the inherent constraints of auto-committing might lead to:
```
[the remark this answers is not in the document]
ppatierno: @tombentley you are changing with this:

> When enabled, consumers commit the offsets of the just-processed messages automatically when `poll()` is called to get the next messages to process

it sounds that if my application has auto commit every 3 seconds but it does a poll every 30 seconds, the offsets will not be committed after 3 seconds but it will wait 30 seconds (the next poll). Aren't these two things not related? I mean, isn't auto commit async compared to the poll? Otherwise the `auto.commit.interval.ms` doesn't make so much sense, or?

_posts/2021-01-08-consumer-tuning.md
```diff
@@ -0,0 +1,292 @@
...
+
+If this potential situation leaves you slightly concerned, what can you do about it? First of all, you can use the `auto.commit.interval.ms` property to decrease those worrying intervals between commits. But this will not completely eliminate the chance that messages are lost or duplicated.
+
+What else? Alternatively, you can turn off auto-committing by setting `enable.auto.commit` to `false`. You then assume responsibility for how your consumer application handles commits correctly. This is an important decision. Periodic automatic offset commits do mean it's something you don't need to worry about. And as long as all message processing is done before the next poll, all processed offsets will be committed. But a higher level of control might be preferable if data loss or data duplication is to be avoided.   
+
+```properties
+# ...
+enable.auto.commit=false
+auto.commit.interval.ms=1000
+# ...
```
ppatierno, line 158: I would move the `auto.commit.interval.ms=1000` example at the end of the paragraph where we mention it and leave the auto commit disabling here, right after the paragraph mentioning it.
Having this example with both changes could be misleading (I am disabling but reducing the commit interval??)

_posts/2021-01-08-consumer-tuning.md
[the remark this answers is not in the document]
tombentley: @ppatierno ah, you're quite right! I do apologise. Is this better:

```suggestion
Kafka's _auto-commit mechanism_ is pretty convenient (and sometimes suitable, depending on the use case). When enabled, consumers commit the offsets of messages automatically every `auto.commit.interval.ms` milliseconds. But convenience, as always, has a price. By allowing your consumer to commit offsets, you are introducing a risk of data loss and duplication. How? 

* If your application commits an offset and then crashes before all the messages up to that offset have actually been processed those messages won't get processed when the application is restarted (data loss).
* If your application has processed all the messages and then crashes before the offset is committed automatically then the last offset will be used when the application restarts and you'll process those messages again. 

Turning off the auto commit functionality can help with the first issue, because you can write your code on only commit offsets once messages have actually been processed. But manual commit can't completely eliminate the second problem because you can't guarantee that the offset commit message will always be processed by the broker. This is the case even if you do a synchronous offset commit after every processing each message.
```

?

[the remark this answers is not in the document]
PaulRMellor: I updated the first bit as follows:

image

I added the final paragraph to sum up the section:

image

_posts/2021-01-08-consumer-tuning.md
PaulRMellor: Good idea. I've separated.

ppatierno (approved):

PR #220, opened 2021-03-16
Remove OKD from the Quickstarts

scholzj:
The Quickstarts still have a quick start guide for OKD which uses `oc cluster up`. But we do not support such old Kube / OCP versions. This PR comments out the OKD section so that we can replace it easily with something new in the future. Possibly with CRC based OKD if it works for anyone? Or maybe some cloud providers etc.


Frawless (approved):
I think we should also remove oc cluster up/minishift from docs like here https://strimzi.io/docs/operators/latest/deploying.html#deploy-alternatives_str

ppatierno (approved):

PR #231, opened 2021-04-15
Add link to full docs

scholzj:
Currently, the website has two versions of the documentation: one is embedded into the website it self:
* for example https://strimzi.io/docs/operators/latest/quickstart.html

Other one is using the default Asciidoc rendered:
* for example https://strimzi.io/docs/operators/latest/full/quickstart.html

To get to the other one, you need to know how to change the URL. But in some cases, the full documentation is more readable. So this PR adds new icons to the documentation website which can be used to directly open the _full_ documentation in a new window.

Screenshot 2021-04-16 at 0 14 20

@strimzi/maintainers please have a look at this and let me know what do you think about it. You can also try it in the preview.

### DIFF
--- a/_sass/core/includes/secondary-page-title-band.scss
+++ b/_sass/core/includes/secondary-page-title-band.scss
@@ -1,5 +1,5 @@
 .secondary-page-title-band {
-  .documentation-verion-band {
+  .documentation-version-band {
     i {
       color: $white !important;
       text-decoration: none;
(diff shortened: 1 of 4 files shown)


Frawless (approved):

_sass/core/includes/secondary-page-title-band.scss
```diff
@@ -1,3 +1,8 @@
 .secondary-page-title-band {
-
+  .documentation-verion-band {
```
sknot-rh, line 2: If I wanted to be pedantic, I would tell you that there is a typo.

sknot-rh (approved):

PR #240, opened 2021-05-20
[POST] broker configuration tuning

PaulRMellor:

New blog post to describe broker tuning


PaulRMellor:
Taken from the content added to the _Using Guide_. 
Edited and made a little more conversational. Images in Strimzi web colors.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,441 @@
...
+
+And also increase the maximum number of bytes Kafka can receive:
+
+```properties
+# ...
+socket.request.max.bytes=104857600
+# ...
+```
+
+### Increasing bandwidth for high latency connections
```
scholzj, line 158: Would _throughput_ be better than _bandwidth_ here?

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,441 @@
...
+
+If you've fine-tuned the size of your message batches, but latency is impeding the performance of your Kafka environment, try increasing the size of the buffers for sending and receiving messages.
```
scholzj, line 160: Not sure I understand what is this trying to say. Increasing buffers normally improves throughput at the cost of higher latency. Is that the case here? Maybe we can make it a bit more clear.

scholzj (commented):
* Can you set the data to today or yesterday? That way, the blog post will also show in the preview and we can check the formatting etc.
* The examples are confusing. When reading the post they look like recommendations to me ... but they are really just the default values.

PaulRMellor:
> * Can you set the data to today or yesterday? That way, the blog post will also show in the preview and we can check the formatting etc.
> * The examples are confusing. When reading the post they look like recommendations to me ... but they are really just the default values.

I've changed the date.
For the example values, I've added to the intro: _Where example values are shown for properties, this is usually the default — adjust accordingly._

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,441 @@
+---
+layout: post
+title:  "Optimizing Kafka brokers"
```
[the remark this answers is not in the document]
PaulRMellor: I was going with the title format we used for producers and consumers. I've changed here though.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,441 @@
...
+Base the value on the average message size and throughput. When considering the total memory allocation required for read/write buffering, the memory available must also be able to accommodate the maximum replicated message size when multiplied by all followers. The size must also be greater than `message.max.bytes` so that all messages can be replicated.
+
+```properties
+# ...
+replica.fetch.max.bytes=1048576
+# ...
+```
+
+The importance of Kafka's topic replication mechanism cannot be overstated.
+Topic replication is central to Kafka's reliability and data durability. Using replication, a failed broker can recover from its in-sync replicas.
```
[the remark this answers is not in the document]
PaulRMellor: I can see how that can be misinterpreted. Changed.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,441 @@
...
+default.replication.factor=3
+min.insync.replicas=2
+# ...
+```
+
+High availability environments require a replication factor of at least 3 for topics and a minimum number of in-sync replicas as 1 less than the replication factor.
+For increased data durability, set `min.insync.replicas` in your *topic* configuration and message delivery acknowledgments using `acks=all` in your *producer* configuration.
+
+Use the `replica.fetch.max.bytes` property to set a maximum size for the messages fetched by each _follower_ that replicates the data from a _leader_ partition.
+Base the value on the average message size and throughput. When considering the total memory allocation required for read/write buffering, the memory available must also be able to accommodate the maximum replicated message size when multiplied by all followers. The size must also be greater than `message.max.bytes` so that all messages can be replicated.
```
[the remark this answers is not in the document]
PaulRMellor: I changed to: _The size must also be greater than `message.max.bytes`, otherwise the broker can accept messages that are too big to replicate._

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,441 @@
...
+# ...
+replica.fetch.max.bytes=1048576
+# ...
+```
+
+The importance of Kafka's topic replication mechanism cannot be overstated.
+Topic replication is central to Kafka's reliability and data durability. Using replication, a failed broker can recover from its in-sync replicas.
+We go into more detail about leaders, followers and in-sync replicas with [partition rebalancing for availability](#partition-rebalancing-for-availability).
+
+The `auto.create.topics.enable` property to create topics automatically is usually disabled. Kafka users tend to prefer applying more control over topic creation.
```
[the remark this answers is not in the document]
PaulRMellor: To make it clearer, I've changed to: 

> The `auto.create.topics.enable` property is enabled by default to create topics automatically, but it's usually disabled as Kafka users tend to prefer applying more control over topic creation.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,441 @@
...
+If you do use automatic topic creation, set `num.partitions` to equal the number of brokers in the cluster so that writes are distributed.
+The `delete.topic.enable` property to allow topics to be deleted is usually disabled too. This time you're making sure that data is not accidentally lost, although you can temporarily enable the property to delete topics if circumstances demand it.
```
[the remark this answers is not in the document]
PaulRMellor: Changed to:

> The `delete.topic.enable` property is enabled by default to allow topics to be deleted and is usually disabled too.

_posts/2021-05-06-broker-tuning.md
PaulRMellor: Not sure. We talk largely about _bandwidth delay_ in this section.

_posts/2021-05-06-broker-tuning.md
PaulRMellor: Changed to: 

> If you've fine-tuned the size of your message batches, the default values of the buffers for sending and receiving messages might be too small for the required throughput.

_posts/2021-05-06-broker-tuning.md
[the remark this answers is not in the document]
scholzj: TBH, I do not think this makes it clear to me ... the _but it's usually disabled_ makes no sense if it is enabled by default because it will not disable it self. I think it should say for example _but users usually disable it_ to make it clear that it is the user's responsibility to disable it.

_posts/2021-05-06-broker-tuning.md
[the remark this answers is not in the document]
scholzj: Same as above ...

> TBH, I do not think this makes it clear to me ... the _but it's usually disabled_ makes no sense if it is enabled by default because it will not disable it self. I think it should say for example _but users usually disable it_ to make it clear that it is the user's responsibility to disable it.

_posts/2021-05-06-broker-tuning.md
scholzj: I'm not sure readers will know what `bandwidth-delay` really is. But that is fine since you added the Wiki link explaining it. But do you think the readers will reads this section because they think about _bandwidth issues_? Or will they think about _throughput issues_? The blog post is quite long, only few people will read it start to end. Most of them will start reading the titles and read the parts they see as relevant.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+***:warning: Do not reduce these settings in production.***
+
+### Improving request handling throughput by increasing I/O threads
+
+Network threads (`num.network.threads`) handle requests to the Kafka cluster, such as produce and fetch requests from client applications.
+Adjust the number of network threads to reflect the replication factor and the levels of activity from client producers and consumers interacting with the Kafka cluster.
+
+I/O threads (`num.io.threads`) pick up requests from the request queue to process them. Adding more threads can improve throughput, but the number of CPU cores/and disk bandwidth imposes a practical upper limit.
+The minimum number of threads should equal the number of storage volumes.
```
tombentley, line 127: I wonder if we're phrasing this right. It's true that that is a reasonable minimum, but in practice people don't want to scale this down to a minimum, they want to tune it to what's best for their disks, and will happily turn it up. The default is 8, so maybe suggesting that `8*#disks` is a reasonable place to start tuning is more useful that mentioning a minimum.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+```properties
+# ...
+num.io.threads=8
+queued.max.requests=500
+num.network.threads=3
+num.recovery.threads.per.data.dir=1
+# ...
+```
+
```
tombentley, line 140: Since we've covered most of the thread pools here, it might also be worth mentioning that when reconfiguring these things """The size of various thread pools used by the broker may be updated dynamically at cluster-default level used by all brokers. Updates are restricted to the range currentSize / 2 to currentSize * 2 to ensure that config updates are handled gracefully. """

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+You can work out the optimal size of your buffers using a _bandwidth-delay_ product calculation.
+You need some figures to work with. Essentially, you want to multiply your bandwidth capacity with the round-trip delay to give the data volume.
+Go look at this [wiki definition](https://en.wikipedia.org/wiki/Bandwidth-delay_product) to see examples of the calculation.
+
+### Managing logs with data retention policies
+
+Kafka uses logs to store message data. Logs are a series of segments. New messages are written to a single _active_ segment. The messages are never modified.
```
tombentley, line 177:
```suggestion
Kafka uses logs to store message data. Logs are a series of segment files with various associated indexes. New messages are written to a single _active_ segment. The messages are never modified.  
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+```
+
+You can work out the optimal size of your buffers using a _bandwidth-delay_ product calculation.
+You need some figures to work with. Essentially, you want to multiply your bandwidth capacity with the round-trip delay to give the data volume.
+Go look at this [wiki definition](https://en.wikipedia.org/wiki/Bandwidth-delay_product) to see examples of the calculation.
+
+### Managing logs with data retention policies
+
+Kafka uses logs to store message data. Logs are a series of segments. New messages are written to a single _active_ segment. The messages are never modified.  
+Using fetch requests, consumers read from the active segment. After a time, the active segment is _rolled_ to become ready-only as a new active segment replaces it.
```
tombentley, line 178: Consumers don't always ready from the active segment, they can read from any.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+Larger sizes means the active segment contains more messages and is rolled less often.
+Non-active segments also become eligible for deletion less often.
+
+You can set time-based or size-based log retention policies, in conjunction with the cleanup policies we cover next, so that logs are kept manageable.
+Non-active log segments are removed when retention limits are reached.
+By controlling the size of the log you retain, you can make sure that you're not likely to exceed disk capacity.
+
+For time-based log retention, use the milliseconds configuration, which has priority over the related minutes and hours configuration.
+The milliseconds configuration also updates dynamically.
```
tombentley, line 199:
```suggestion
The milliseconds configuration also support dynamic update, which the other configurations do not.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+```properties
+# ...
+log.retention.ms=1680000
+# ...
+```
+
+The retention period is based on the time messages were appended to the segment.
+If  `log.retention.ms` is set to -1, no time limit is applied to log retention, so all logs are retained.  
+Disk usage should always be monitored, but the -1 setting is not generally recommended as it can lead to issues with full disks, which can be hard to rectify.
```
tombentley, line 209:
```suggestion
Disk usage should always be monitored, but the -1 setting is not generally recommended as it is especially likely to lead to issues with full disks, which can be hard to rectify.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+For size-based log retention, set a maximum log size in bytes:
+
+```properties
+# ...
+log.retention.bytes=1073741824
+# ...
+```
+
+The maximum log size is for all segments in the log.
```
tombentley, line 219:
```suggestion
The maximum log size is for all segments in the log. In other words a log will typically end up with approximately `retention.bytes/segment.bytes` segments once it reaches a steady state.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+# ...
+log.cleanup.policy=compact,delete
+# ...
+```
+
+* **Delete policy**
+`delete` policy corresponds to managing logs with data retention policies. It's suitable when data does not need to be retained forever. Older segments are deleted based on log retention limits. Otherwise, if the log cleaner is not enabled, and there are no log retention limits, the log will continue to grow.
+
+* **Compact policy**
+`compact` policy guarantees to keep the most recent message for each message key. Log compaction is suitable when message values are changeable, and you want to retain the latest update. New messages are appended to the  _head_ of the log, which acts in the same way as a non-compacted log with writes appended in order. The _tail_ of a compacted log has the older messages that are deleted or compacted according to policy. Consequently, the tail has non-contiguous offsets.
```
tombentley, line 241:
```suggestion
`compact` policy guarantees to keep the most recent message for each message key. Log compaction is suitable when message values are changeable, and you want to retain the latest update. New messages are appended to the  _head_ of the log, which acts in the same way as a non-compacted log with writes appended in order. The _tail_ of a compacted log has the older messages that are deleted or compacted according to policy. Consequently, the tail has non-contiguous offsets. While Kafka guarantees that the latest messages for each key will be retained, it does not guarantee that the whole compacted log will not contain duplicates.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+**Log showing key value writes with offset positions before compaction**
+
+![Image of compaction showing key value writes](/assets/images/posts/2021-05-06-broker-tuning-compaction-1.png)
+
+If you're not using keys, you can't use compaction as keys are needed to identify related messages. The latest message (with the highest offset) is kept and older messages with the same key are discarded. You can restore a message back to a previous state. Records retain their original offsets after cleanup. When consuming an offset that's no longer available in the tail, the record with the next higher offset is found.
```
tombentley, line 247: IDK what "You can restore a message back to a previous state." is supposed to mean

```suggestion
If you're not using keys, you can't use compaction as keys are needed to identify related messages. During log cleaning the latest message (with the highest offset) is kept and older messages with the same key are discarded.  Records retain their original offsets after cleanup. When consuming an offset that's no longer available in the tail, the record with the next higher offset is found.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+After the log has been cleaned up, compacted messages are no longer available in the tail of the log. Records retain their original offset. Consuming from the offset of record that is no longer available returns the next higher available offset.
```
tombentley, line 249: Repetition with the previous para

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+#### Reference-based messaging
+Reference-based messaging only sends a reference to data stored in some other system in the message’s value. This is useful for data replication when you do not know how big a message will be.
+Data is written to the data store, which must be fast, durable, and highly available, and a reference to the data is returned. The producer sends the reference to Kafka. The consumer uses the reference to fetch the data from the data store.
+
+**Reference-based messaging flow**
+
+![Image of reference-based messaging flow](/assets/images/posts/2021-05-06-broker-tuning-reference-messaging.png)
+
+Referenced-based messaging means more for message passing requires more trips, so end-to-end latency will increase. One major drawback of this approach is that there is no automatic clean up of the data in the external system when the Kafka message is cleaned up.
```
tombentley, line 332: Slight garbage in the first sentence?
```suggestion
Referenced-based messaging means more for message passing requires more trips, so end-to-end latency will increase. One major drawback of this approach is that there is no automatic clean up of the data in the external system when the Kafka message is cleaned up. There is also the risk that data could be removed from the external system before the message is Kafka gets deleted.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+One way to mitigate the limitations of reference-based messaging is to take a hybrid approach. Send large messages to the data store and process standard-sized messages directly.
+
+#### Inline messaging
+Inline messaging is a complex process that splits messages into chunks that use the same key, which are then combined on output using a stream processor like Kafka Streams.
+
+The basic steps are:
+
+- The client application serializes then chunks the data if the message is too big.
```
tombentley, line 341: Let's try to be clear what is done for you by the producer and consumer v.s. what you have to do for yourself in the application logic. 

```suggestion
- The producing client application serializes then chunks the data if the message is too big.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+- The producer uses the Kafka `ByteArraySerializer` or similar to serialize each chunk again before sending it.
+- The consumer receives the chunks, which are assembled before deserialization.
```
tombentley, line 343:
```suggestion
- The consuming client application receives the chunks, which are assembled before deserialization.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+- The consumer tracks messages and buffers chunks until it has a complete message.
```
tombentley, line 344: Perhaps this one should be swapped with the previous bullet?

```suggestion
- The consuming client application tracks messages and buffers chunks until it has a complete message.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+- Complete messages are delivered in order according to the offset of the first or last chunk for each set of chunked messages.
```
tombentley, line 345:
```suggestion
- Complete messages are delivered to the rest of the consuming application in order according to the offset of the first or last chunk for each set of chunked messages.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+#### Configuration to handle larger messages**
+
+You increase message limits by configuring `message.max.bytes` to set the maximum record batch size. You can set the limit at the topic level or the broker level. If you set limit at the broker level, larger messages are allowed for all topics and messages greater than the maximum limit are rejected. The buffer size for the producers (`max.request.size`) and consumers (`message.max.bytes`) must be able to accommodate the larger messages.
+
+### Controlling the log flush of message data
+
+Usually, log flush thresholds are not usually set and the operating system performs a background flush using its default settings. But if you are using application flush management, setting lower flush thresholds might be appropriate if you are looking at ways to decrease latency or you are using faster disks.
```
tombentley, line 360: Flushing isn't going to decrease latency.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+You can adjust the lag time before a follower is considered out of sync:
+
+```properties
+# ...
+replica.lag.time.max.ms
+# ...
+```
+
+This is the time to replicate a message to all in-sync replicas and return an acknowledgment to a producer. If a follower fails to make a fetch request and catch up with the latest message within the specified lag time, it is removed from in-sync replicas. The time you use depends on both network latency and broker disk bandwidth. You can reduce the lag time to detect failed replicas sooner, but by doing so you might increase the number of followers that fall out of sync needlessly.
```
tombentley, line 388:
```suggestion
This is an upper bound on the time to replicate a message to all in-sync replicas and return an acknowledgment to a producer. If a follower fails to make a fetch request and catch up with the latest message within the specified lag time, it is removed from in-sync replicas. The time you use depends on both network latency and broker disk bandwidth. You can reduce the lag time to detect failed replicas sooner, but by doing so you increase the chances of followers falling out of sync needlessly.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+>Followers operate only to replicate messages from the partition leader and allow recovery should the leader fail. Followers do not normally serve clients, though [rack configuration](https://strimzi.io/docs/operators/latest/using.html#type-Rack-reference) allows a consumer to consume messages from the closest replica when a Kafka cluster spans multiple datacenters.
+
+A failed leader affects the balance of a Kafka cluster, as does the assignment of partition replicas to brokers. Rebalancing ensures that leaders are evenly distributed across brokers and brokers are not overloaded. You can control the frequency, in seconds, of rebalance checks and the maximum percentage of imbalance allowed for a broker before a rebalance is triggered.
```
tombentley, line 392: The concept we're trying to introduce here is that of _leader rebalancing_. It's a distinct thing from "_assignment rebalancing_" (which is what CC does), so we should try to be clear about the difference between the two.

```suggestion
A failed leader affects the balance of a Kafka cluster, as does the assignment of partition replicas to brokers. Leader rebalancing ensures that leaders are evenly distributed across brokers and brokers are not overloaded. You can control the frequency, in seconds, of leader rebalance checks and the maximum percentage of imbalance allowed for a broker before a rebalance is triggered.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+```properties
+#...
+auto.leader.rebalance.enable=true
+leader.imbalance.check.interval.seconds=300
+leader.imbalance.per.broker.percentage=10
+#...
+```
+
+The percentage imbalance for a broker is the gap between the current number of partition leaders it holds and the number of partitions which are preferred leaders. The first broker in a partition’s list of replicas is known as the preferred leader. You can set the percentage to zero to ensure that preferred leaders are always elected.
```
tombentley, line 402:
```suggestion
The first broker in a partition’s list of replicas is known as the preferred leader. The percentage leader imbalance for a broker is the ratio between the current number of partitions for which is the _current_ leader and the number of partitions for which it is the _preferred_ leader.  You can set the percentage to zero to ensure that preferred leaders are always elected (assuming they're in the ISR).
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+If the checks for rebalances need more control, you can disable automated rebalances. You can then choose when to trigger a rebalance using the `kafka-leader-election.sh` command line tool.
+Alternatively, you can [use Cruise Control for Strimzi](https://strimzi.io/docs/operators/latest/using.html#cruise-control-concepts-str) to change partition leadership and rebalance replicas across your Kafka cluster in a more intelligent way.
+
+>The Grafana dashboards provided with Strimzi show metrics for under-replicated partitions and partitions that do not have an active leader.
+
+### Unclean leader election
+
+Leader election to an in-sync replica is considered _clean_ because it guarantees no loss of data. And this is what happens by default. Kafka waits until the original leader is back online before messages are picked up again.
```
tombentley, line 411:
```suggestion
Leader election to an in-sync replica is considered _clean_ because it guarantees no loss of data. And this is what happens by default. If no other broker was in the ISR when the old leader was lost, Kafka waits until that leader is back online before messages can be written or read.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+But what if there is no in-sync replica to take on leadership? If a minimum number of in-sync replicas is not set, and there are no followers in sync with the partition leader when its hard drive fails irrevocably, data is already lost. Not only that, but a new leader cannot be elected because there are no in-sync followers.
```
tombentley, line 413:
```suggestion
But what if there is no in-sync replica to take on leadership, for example because the ISR only contained the leader when the leader's disk died? If a minimum number of in-sync replicas is not set, and there are no followers in sync with the partition leader when its hard drive fails irrevocably, data is already lost. Not only that, but _a new leader cannot be elected_ because there are no in-sync followers.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+# ...
+unclean.leader.election.enable=false
+# ...
+```
+
+Unclean leader election means out-of-sync replicas can become leaders, but you risk losing messages.
+
+### Avoiding unnecessary consumer group rebalances
+
+We''l end this exploration of broker tuning tips by suggesting a useful configuration to avoid unnecessary consumer group rebalances. You can add a delay so the group coordinator waits for members to join before the initial rebalance.
```
tombentley, line 427:
```suggestion
We'll end this exploration of broker tuning tips by suggesting a useful configuration to avoid unnecessary consumer group rebalances. You can add a delay so the group coordinator waits for members to join before the initial rebalance.
```

_posts/2021-05-06-broker-tuning.md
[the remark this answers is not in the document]
PaulRMellor: Right. I get what you're trying to say. Changed to:

> The `auto.create.topics.enable` property is enabled by default to create topics automatically, but Kafka users tend to disable it so they can apply more control over topic creation.

_posts/2021-05-06-broker-tuning.md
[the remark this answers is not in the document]
PaulRMellor: Changed to:

> The `delete.topic.enable` property is enabled by default to allow topics to be deleted. Kafka users normally disable this property too. This time you're making sure that data is not accidentally lost, although you can temporarily enable the property to delete topics if circumstances demand it.

_posts/2021-05-06-broker-tuning.md
PaulRMellor: I see your point. I've changed to _throughput_ in the title.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+---
+
+Following our popular and evergreen posts on fine-tuning [producers](https://strimzi.io/blog/2020/10/15/producer-tuning/) and [consumers](https://strimzi.io/blog/2021/01/07/consumer-tuning/),
+we complete our _Kafka optimization trilogy_ with this post on getting started with broker configuration.
+
+In terms of the options you can use to configure your brokers, there are countless permutations.
+We've reduced and refined the options and present here what's typically configured to get the most out of Kafka brokers.
+
+And when we say brokers, we also include topics.
+Broker configuration often corresponds with configuration options at the topic level, such as setting the maximum batch size for topics with `message.max.bytes`.
```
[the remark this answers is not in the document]
PaulRMellor: I've updated and removed the second sentence.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+replica.fetch.max.bytes=1048576
+# ...
+```
+
+The importance of Kafka's topic replication mechanism cannot be overstated.
+Topic replication is central to Kafka's reliability and data durability. Using replication, a failed broker can recover from in-sync replicas on other brokers.
+We go into more detail about leaders, followers and in-sync replicas with [partition rebalancing for availability](#partition-rebalancing-for-availability).
+
+The `auto.create.topics.enable` property is enabled by default to create topics automatically, but it's usually disabled as Kafka users tend to prefer applying more control over topic creation.
+If you do use automatic topic creation, set `num.partitions` to equal the number of brokers in the cluster so that writes are distributed.
```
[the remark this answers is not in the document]
PaulRMellor: This section fell out of sync with the content in the main doc. I've changed to:

> If you are using automatic topic creation, you can set the default number of partitions for topics using `num.partitions`. Use it with `default.replication.factor` property. In this case, you might want to set the replication factor to the number of brokers or to at least three replicas.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+The `delete.topic.enable` property is enabled by default to allow topics to be deleted and is usually disabled too. This time you're making sure that data is not accidentally lost, although you can temporarily enable the property to delete topics if circumstances demand it.
+
+```properties
+# ...
+auto.create.topics.enable=false
+delete.topic.enable=true
+# ...
+```
+
+> You cannot delete topics with the `KafkaTopic` resource if the `delete.topic.enable` property is set to false.
```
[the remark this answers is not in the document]
PaulRMellor: I've added:

> If you wish to add a time delay before a segment file is deleted from the system, you can add the delay using `log.segment.delete.delay.ms` for all topics in the broker configuration or `file.delete.delay.ms` for specific topics in the topic configuration. 

I thought this was better in the _Removing log data with cleanup policies_ section

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,443 @@
...
+
+### Internal topic settings for transactions and commits
+
+Pay attention to the configuration requirements for internal Kafka topics.
+
+If you are [using transactions](https://strimzi.io/docs/operators/latest/using.html#reliability_guarantees) to guarantee the reliability of writes to partitions from producers, the internal `__transaction_state` topic used in the process requires a minimum of three brokers with the default settings.
```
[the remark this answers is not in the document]
PaulRMellor: Changed to:

> If you are [using transactions](https://strimzi.io/docs/operators/latest/using.html#reliability_guarantees) to enable atomic writes to partitions from producers, the internal `__transaction_state` topic used in the process requires a minimum of three brokers with the default settings.

_posts/2021-05-06-broker-tuning.md
PaulRMellor: Makes sense. I've changed to:

> A good starting point might be to start with the default of 8 multiplied by the number of disks.

_posts/2021-05-06-broker-tuning.md
PaulRMellor: Added:

> Configuration updates to the thread pools for all brokers might occur dynamically at the cluster level. 
> These updates are restricted to between half the current size and twice the current size.

_posts/2021-05-06-broker-tuning.md
PaulRMellor: Changed to:

> Using fetch requests, consumers read from the segments.

_posts/2021-05-06-broker-tuning.md
PaulRMellor: I've removed the paragraph

_posts/2021-05-06-broker-tuning.md
PaulRMellor: Removed

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+
+```properties
+# ...
+offsets.topic.num.partitions=50
+offsets.topic.replication.factor=3
+offsets.commit.required.acks=-1
+# ...
+```
+
+***:warning: Do not reduce these settings in production.***
```
scholzj, line 120: This does not seem to render on the website ... it just shows `:warning: Do not reduce these settings in production.` in bold. Maybe we can do something like:

```suggestion
***WARNING: Do not reduce these settings in production.***
```

scholzj (approved):
One nit. But LGTM otherwise.

PaulRMellor:
@tombentley Hi Tom. Any more comments?

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+log.flush.interval.ms=50000
+log.flush.interval.messages=100000
+# ...
+```
+
+The wait between flushes includes the time to make the check and the specified interval before the flush is carried out. Increasing the frequency of flushes can affect throughput.
+
+### Partition rebalancing for availability
+
+When replicating data across brokers, a partition leader on one broker handles all producer requests (writes to the log). Partition followers on other brokers replicate the partition data of the leader. So you get data reliability in the event of the leader failing. Followers need to be in sync for recovery, meaning the follower has caught up with the most recently committed message on the leader. If a leader is no longer available, one of the in-sync replicas is chosen as the new leader. The leader checks the follower by looking at the last offset requested.
```
tombentley, line 385:
```suggestion
When replicating data across brokers, a partition leader on one broker handles all produce requests (writes to the log). Partition followers on other brokers replicate the partition data of the leader. So you get data reliability in the event of the leader failing. Followers need to be in sync for recovery, meaning the follower has caught up with the most recently committed message on the leader. If a leader is no longer available, one of the in-sync replicas is chosen as the new leader. The leader checks the follower by looking at the last offset requested.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+
+>Followers operate only to replicate messages from the partition leader and allow recovery should the leader fail. Followers do not normally serve clients, though [rack configuration](https://strimzi.io/docs/operators/latest/using.html#type-Rack-reference) allows a consumer to consume messages from the closest replica when a Kafka cluster spans multiple datacenters.
+
+A failed leader affects the balance of a Kafka cluster, as does the assignment of partition replicas to brokers. Leader rebalancing ensures that leaders are evenly distributed across brokers and brokers are not overloaded. You can control the frequency, in seconds, of leader rebalance checks and the maximum percentage of imbalance allowed for a broker before a rebalance is triggered.
```
tombentley, line 401:
```suggestion
You can [use Cruise Control for Strimzi](https://strimzi.io/docs/operators/latest/using.html#cruise-control-concepts-str) to figure out replica assignments to brokers that balance load evenly across the cluster. It's calculation takes into account the differing load experienced by leaders and followers. A failed leader affects the balance of a Kafka cluster because the remaining brokers get the extra work of leading additional partitions. 

For the assignment found by Cruise Control to actually be balanced it is necessary that partitions are lead by the preferred leader. Kafka can automatically ensure that the preferred leader is being used (where possible), changing the current leader if necessary. This ensures that the cluster remains in the balanced state found by Cruise Control.

You can control the frequency, in seconds, of automatic preferred leader checks and the maximum percentage of imbalance allowed for a broker before a rebalance is triggered.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+
+```properties
+#...
+auto.leader.rebalance.enable=true
+leader.imbalance.check.interval.seconds=300
+leader.imbalance.per.broker.percentage=10
+#...
+```
+
+The first broker in a partition’s list of replicas is known as the preferred leader.
```
tombentley, line 411:
```suggestion
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+The percentage leader imbalance for a broker is the ratio between the current number of partitions for which the broker is the _current_ leader and the number of partitions for which it is the _preferred_ leader.  
+You can set the percentage to zero to ensure that preferred leaders are always elected, assuming they are in sync.
+
+If the checks for rebalances need more control, you can disable automated rebalances. You can then choose when to trigger a rebalance using the `kafka-leader-election.sh` command line tool.
+Alternatively, you can [use Cruise Control for Strimzi](https://strimzi.io/docs/operators/latest/using.html#cruise-control-concepts-str) to change partition leadership and rebalance replicas across your Kafka cluster in a more intelligent way.
```
tombentley, line 416:
```suggestion
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+default.replication.factor=3
+min.insync.replicas=2
+# ...
+```
+
+High availability environments require a replication factor of at least 3 for topics and a minimum number of in-sync replicas as 1 less than the replication factor.
+For increased data durability, set `min.insync.replicas` in your *topic* configuration and message delivery acknowledgments using `acks=all` in your *producer* configuration.
+
+Use the `replica.fetch.max.bytes` property to set a maximum size for the messages fetched by each _follower_ that replicates the data from a _leader_ partition.
+Base the value on the average message size and throughput. When considering the total memory allocation required for read/write buffering, the memory available must also be able to accommodate the maximum replicated message size when multiplied by all followers. The size must also be greater than `message.max.bytes`, otherwise the broker can accept messages that are too big to replicate.
```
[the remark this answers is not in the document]
PaulRMellor: Removed

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+
+### Improving request handling throughput by increasing I/O threads
+
+Network threads (`num.network.threads`) handle requests to the Kafka cluster, such as produce and fetch requests from client applications.
+Adjust the number of network threads to reflect the replication factor and the levels of activity from client producers and consumers interacting with the Kafka cluster.
+
+I/O threads (`num.io.threads`) pick up requests from the request queue to process them. Adding more threads can improve throughput, but the number of CPU cores/and disk bandwidth imposes a practical upper limit.
+A good starting point might be to start with the default of 8 multiplied by the number of disks.
+
+Try starting with the  minimum number of threads should equal the number of storage volumes.
```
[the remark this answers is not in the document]
PaulRMellor: Legacy text! Removed.

_posts/2021-05-06-broker-tuning.md
```diff
@@ -0,0 +1,454 @@
...
+queued.max.requests=500
+num.network.threads=3
+num.recovery.threads.per.data.dir=1
+# ...
+```
+
+Configuration updates to the thread pools for all brokers might occur dynamically at the cluster level.
+These updates are restricted to between half the current size and twice the current size.  
+
+> Kafka broker metrics can help with working out the number of threads required. For example, metrics for the average time network threads are idle (`kafka.network:type=SocketServer,name=NetworkProcessorAvgIdlePercent`) indicate the percentage of resources used.
```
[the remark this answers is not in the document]
PaulRMellor: Moved up

_posts/2021-05-06-broker-tuning.md
```diff
@@ -122,14 +123,14 @@ offsets.commit.required.acks=-1
...
+Adjust the number of network threads to reflect the replication factor and the levels of activity from client producers and consumers interacting with the Kafka cluster. To reduce congestion and regulate the request traffic, you can use `queued.max.requests` to limit the number of requests allowed in the request queue before the network thread is blocked.
 
-I/O threads (`num.io.threads`) pick up requests from the request queue to process them. Adding more threads can improve throughput, but the number of CPU cores/and disk bandwidth imposes a practical upper limit.
-A good starting point might be to start with the default of 8 multiplied by the number of disks.
+> Kafka broker metrics can help with working out the number of threads required. For example, metrics for the average time network threads are idle (`kafka.network:type=SocketServer,name=NetworkProcessorAvgIdlePercent`) indicate the percentage of resources used.
+If there is 0% idle time, all resources are in use, which means that more threads would be beneficial.
```
tombentley, line 129:
```suggestion
If there is 0% idle time, all resources are in use, which means that more threads could be beneficial.
```

_posts/2021-05-06-broker-tuning.md
```diff
@@ -398,7 +402,11 @@ Lag time puts an upper limit on the time to replicate a message to all in-sync r
...
 
-A failed leader affects the balance of a Kafka cluster, as does the assignment of partition replicas to brokers. Leader rebalancing ensures that leaders are evenly distributed across brokers and brokers are not overloaded. You can control the frequency, in seconds, of leader rebalance checks and the maximum percentage of imbalance allowed for a broker before a rebalance is triggered.
+You can [use Cruise Control for Strimzi](https://strimzi.io/docs/operators/latest/using.html#cruise-control-concepts-str) to figure out replica assignments to brokers that balance load evenly across the cluster. It's calculation takes into account the differing load experienced by leaders and followers. A failed leader affects the balance of a Kafka cluster because the remaining brokers get the extra work of leading additional partitions.
```
tombentley, line 405:
```suggestion
You can [use Cruise Control for Strimzi](https://strimzi.io/docs/operators/latest/using.html#cruise-control-concepts-str) to figure out replica assignments to brokers that balance load evenly across the cluster. Its calculation takes into account the differing load experienced by leaders and followers. A failed leader affects the balance of a Kafka cluster because the remaining brokers get the extra work of leading additional partitions.
```

tombentley (approved):
A couple of nits, otherwise LGTM.

PR #269, opened 2021-12-09
[BLOG POST] Blog post about segments, rolling and retention

ppatierno:
### Type of change

* [ ] Typo/minor fix
* [ ] Other


scholzj (commented):
If you set the data to today or yesterday, we will be able to see the preview -> it will be easier to review it.

_posts/2021-12-13-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
+---
+layout: post
+title: "Deep dive into Apache Kafka storage internals: segments, rolling and retention"
+date: 2021-12-13
+author: paolo_patierno
+---
+
+Everyone who uses Apache Kafka knows that it behaves as a commit-log when it comes to deal with storing records.
```
scholzj: Maybe we can word it without the _everyone knows_ ... it is a bit passive-agressive towards people who might not know ;-)

```suggestion
Apache Kafka behaves as a commit-log when it comes to deal with storing records.
```

ppatierno:
@scholzj done, thanks!

_posts/2021-12-13-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+│   ├── 00000000000000000109.log
+│   ├── 00000000000000000109.snapshot
+│   ├── 00000000000000000109.timeindex
+```
+
+Above, the partition 0 of the topic `__strimzi_canary` used by the Strimzi Canary component; the folder contains the following files:
+
+* The `.log` file is an actual segment containing records up to a specific offset. The name of the file defines the starting offset of the records in that log.
+* The `.index` file contains an index that maps logical offset (in effect the record's id) to the byte offset of the record within the `.log` file. It is used for accessing records at specified offsets in the log without having to scan the whole `.log` file.
+* The `.timeindex` file is another index used for accessing records by timestamp in the log.
```
scholzj: What about the `.snapshot`? You have it both above and below for one of the segments. But you do not seem to explain it.

_posts/2021-12-13-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+In reality, a record could live even longer than our 10 minutes depending on a lot of other configuration and internal mechanics of the broker.
+The usual retention limits set by using `log.retention.ms` defines a kind of lower bound; Kafka guarantees it will not delete any records that have an age less than the one specified, but any older records could get deleted at any time in the future depending on different settings.
+
+It is also worth to mention how it has an impact on the consumer side.
+Consumers don't get records from closed segment as well as from deleted segment even if they are just marked as "deleted" but not actually removed from the file system.
+This is true even when a consumer starts to read the partition from the beginning.
+Longer retention won't have a direct impact on the consumers but more on the disk usage.
+
+It is worth mentioning that the overall retention mechanism works this way when the `log.cleanup.policy` is set to `delete`.
+For `compact`ed topics, it is completely different as you can read about on the official Strimzi documentation [here](https://strimzi.io/docs/operators/latest/using.html#removing_log_data_with_cleanup_policies).
```
scholzj: Maybe this is something you can mention early -> that this applies to topics with `delete` policy and not `compact`. When you do it only at the end, it might confuse the users while reading the post.

scholzj (commented):
Left some comments after a quick pass ... looks good and helpful.

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+In our canary related example, the segment will take 9 minutes to be filled and closed, so when the last record comes, the first one in the segment is already 9 minutes old; while waiting for the 10 minutes retention since the last record, the first record "should be" deleted after 19 minutes.
+
+Anyway, even when we think that finally the retention time is evaluated after the last record, it could be still there!
+The reason is related to an Apache Kafka thread which runs periodically on the broker to clean records and checking which closed segment can be deleted. The `log.retention.check.interval.ms` is the parameter to configure the period for this thread to run and it defaults to 5 minutes.
+Depending on when the last record is appended in the segment, so the segment is closed and when the thread run the last time, it could miss the 10 minutes deadline of the retention and so deleting the closed segment later, so maybe up to almost 5 minutes delay.
```
PaulRMellor, line 180:
```suggestion
Depending on when the last record is appended and a segment is closed, periodic checks for deletion might contribute to missing the 10-minute deadline of the retention period. The closed segment might be deleted on the next check with up to almost 5 minutes delay.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+In our example, it could happen that the first record in the segment could live up to almost 24 minutes!
```
PaulRMellor, line 182:
```suggestion
In our example, it might be that the first record in the segment could live up to almost 24 minutes!
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+Assuming that at some point the clean thread runs and verify that a closed segment can be deleted, initially it just change extensions of corresponding files by adding the `.deleted` extension but not actually deleting the segment from the file system.
```
PaulRMellor, line 184:
```suggestion
Assuming that at some point the clean thread runs and verifies that a closed segment can be deleted. Initially, it adds a `.deleted` extension to the corresponding files, but doesn't actually delete the segment from the file system.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+There is another broker parameter `log.segment.delete.delay.ms` (1 minute by default) which defines when the file will be actually cancelled from the file system when it's marked as "deleted".
```
PaulRMellor, line 185:
```suggestion
The `log.segment.delete.delay.ms` broker parameter defines when the file will actually be removed from the file system when it's marked as "deleted" (1 minute by default).
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+![Total Retention Time](/assets/images/posts/2021-12-09-total-retention-time.png)
+
+Coming back to the canary example again and adding the delay on deletion, the first record in a segment is still alive after 25 minutes!
+It is really longer than the 10 minutes expectation, isn't it?
```
PaulRMellor, line 190:
```suggestion
Returning to the Canary example again and assuming the delay on deletion, the first record in our segment is still live after 25 minutes!
It is really longer than the 10 minutes expectation, isn't it?
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+From the above, it is clear that the retention mechanism doesn't really match the initial expectation.
```
PaulRMellor, line 192:
```suggestion
As you can see, it is clear that the retention mechanism doesn't really match the initial expectation.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+In reality, a record could live even longer than our 10 minutes depending on a lot of other configuration and internal mechanics of the broker.
```
PaulRMellor, line 193:
```suggestion
In reality, a record could live even longer than our 10 minutes depending on the configuration and internal mechanics of the broker.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+The usual retention limits set by using `log.retention.ms` defines a kind of lower bound; Kafka guarantees it will not delete any records that have an age less than the one specified, but any older records could get deleted at any time in the future depending on different settings.
+
+It is also worth to mention how it has an impact on the consumer side.
```
PaulRMellor, line 196:
```suggestion
It is also worth mentioning the impact on the consumer side.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+Consumers don't get records from closed segment as well as from deleted segment even if they are just marked as "deleted" but not actually removed from the file system.
```
PaulRMellor, line 197:
```suggestion
Consumers don't get records from closed segments or deleted segments, even if they are just marked as "deleted" but not actually removed from the file system.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+This is true even when a consumer starts to read the partition from the beginning.
+Longer retention won't have a direct impact on the consumers but more on the disk usage.
```
PaulRMellor, line 199:
```suggestion
Longer retention won't have a direct impact on the consumers, but more on the disk usage.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+It is worth mentioning that the overall retention mechanism works this way when the `log.cleanup.policy` is set to `delete`.
```
PaulRMellor, line 201:
```suggestion
The overall retention mechanism works this way when the `log.cleanup.policy` is set to `delete`.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+For `compact`ed topics, it is completely different as you can read about on the official Strimzi documentation [here](https://strimzi.io/docs/operators/latest/using.html#removing_log_data_with_cleanup_policies).
```
PaulRMellor, line 202:
```suggestion
For compacted (`compact`) topics, it is completely different. You can read more on compaction in the [Strimzi documentation](https://strimzi.io/docs/operators/latest/using.html#removing_log_data_with_cleanup_policies).
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+## Conclusion
+
+Be aware of how the broker stores partitions and corresponding records on the disk is really important.
```
PaulRMellor, line 206:
```suggestion
Understanding how the broker stores partitions and corresponding records on the disk is really important.
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+A lot of configuration parameters can influence how long your data are retained and quite often it's different from your expectation.
```
PaulRMellor, line 207:
```suggestion
Configuration parameters can have a surprisingly big influence on how long your data is retained. 
```

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+Tweaking all these parameters can be really tricky but gives you more power on handling your data.
```
PaulRMellor, line 208:
```suggestion
Learning about these parameters and how you can adjust them gives you a lot more control on how you handle your data.
```

PaulRMellor (commented):
Really interesting read. Anything here we want to include in the docs? 
I've made some suggestions. I have a couple of questions where I'm not sure about a sentence.

_posts/2021-12-13-kafka-segment-retention.md
ppatierno: Good advice, thanks!

_posts/2021-12-13-kafka-segment-retention.md
ppatierno: I am not sure about it ... I mean. How could I start the entire paragraph by saying "everything I am going to explain is valid for delete policy and not for the compact one"? Isn't it better just leaving it out until the end as it is now?
Or at least what could be the better way to say it earlier?

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+How this entries are added inside the index file is defined by the `log.index.interval.bytes` parameter which is 4096 bytes by default.
+It means that every 4096 bytes of records added in the log file, an entry is added in the corresponding index file.
+Each entry is 8 bytes in size, 4 for the offset and 4 for the bytes position in the segment.
+In this example, we have a new index entry added every 28 records because 28 * 150 = 4200.
+
+It means that if a consumer wants to read starting at specific offset, the search of the record will be done with the following steps:
+
+* searching for the `.index` file based on its name which follows the same patterns as the corresponding `.log` file; it contains the starting offset of the records indexed by that index. (TBD: is it a dichotomic/binary search?)
```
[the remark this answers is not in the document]
ppatierno: *its* refers to the index file name that follows the same pattern of the log file name (i.e. 00000000000000.index and 0000000000000.log). *it* refers to the index file itself which contains the starting offset of the records indexed by that index.

_posts/2021-12-13-kafka-segment-retention.md
scholzj: Well, your blog post, your call I guess. But while reading it I was wondering all the time about how does it work with compacting and if you are going to explain it. So it felt a bit weird and disappointing at the end. I thought knowing it upfront would make me think about it a bit less. But as I said, it is your call at the end, so do it as you feel.

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+timestamp: 1638100594371 offset: 84
+```
+
+## Let's talk about rolling segments
+
+A new segment will be rolled when some specific conditions are met.
+One of them is when the segment maximum size specified by the configuration parameter `log.segment.bytes` (1 GiB by default) is reached.
+The second one is based on `log.roll.ms` or `log.roll.hours` (7 days by default) parameters and in this case the segment is rolled when the configured time is elapsed since the producer timestamp of the first record in the segment (or since the creation time if there is no timestamp).
+The first target that is met will cause rolling a new segment.
+It is also worth noticing that, as a not so usual use case, the producer timestamps in the records could be not ordered, so not having the older record as the first one, due to retries or the specific business logic of the producer.
```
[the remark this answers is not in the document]
ppatierno: No, the meaning is different. My sentence wants to highlight that not always the first record in the segment is the older one.  A producer could send a record with older timestamp even if in the middle of a segment.

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+Another useful parameter is `log.roll.jitter.ms` which is used to add a jitter when it's time to roll a segment, to avoid that many segments are rolled at same time causing high contention on the disk.
+A jitter is randomly generated for each segment and the parameter above defines the maximum value.
+
+The above conditions, by size or time, are the well known ones but not everyone knows that there is a third condition.
```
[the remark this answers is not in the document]
ppatierno: The third condition is the one coming in the next sentence.

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,208 @@
...
+
+A segment, together with the records it contains, can be deleted only when it is closed.
+It means that if the producer is pretty slow and the maximum size of 16384 bytes is not reached within the 10 minutes, the older records won't be anyway deleted so our retention will be higher than what we meant to be.
+
+Even if the active segment is filled pretty quickly, the retention time is evaluated starting from the last record appended to the segment itself, right before it is closed.
+It means that the latest record will be retained for the 10 minutes we want but the first records in the segment will be retained more; of course it depends how quickly the segment was filled so how much time passed between the first record and the last one.
+
+In our canary related example, the segment will take 9 minutes to be filled and closed, so when the last record comes, the first one in the segment is already 9 minutes old; while waiting for the 10 minutes retention since the last record, the first record "should be" deleted after 19 minutes.
+
+Anyway, even when we think that finally the retention time is evaluated after the last record, it could be still there!
```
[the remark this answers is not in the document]
ppatierno: I was referring to the first older record but we could refer to all the records in the segment at this point.

_posts/2021-12-13-kafka-segment-retention.md
ppatierno: @PaulRMellor any suggestion where to put this sentence? Tbh I think it's the right place but having a suggestion could help me to understand if it's better somewhere else.

ppatierno:
@PaulRMellor I  addressed all your suggestions and answered your questions. Would be great to have another pass on them if you have further suggestions. Thanks!

scholzj (approved):
Love the graphics!!! 😍

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,259 @@
...
+
+Apache Kafka behaves as a commit-log when it comes to deal with storing records.
+Records are appended at the end of each log one after the other and each log is also split in segments
+Segments help with deletion of older records, improving performance, and much more.
+So, the log is a logical sequence of records that's composed of segments (files) and segments store a sub-sequence of records.
+Broker configuration allows you to tweak parameters related to logs. You can use configuration to control the rolling of segments, record retention, and so on.
+
+Not everyone is aware of how these parameters have an impact on broker behavior.
+For instance, they determine how long records are stored and made available to consumers.
+In this blog post, we will dig more into how the log segment configuration and record retention works, in order to help you make better decisions depending on your particular use case.
```
PaulRMellor, line 16:
```suggestion
In this blog post, we will dig more into how log segmentation and record retention impacts broker performance. When you know more about how it works, you might want to adjust your log configuration. 

> Another aspect of log configuration is enabling compaction. Using keys to identify messages, Kafka compaction keeps the latest message (with the highest offset) for a specific message key. Earlier messages that have the same key are discarded. You can read more on compaction in the Strimzi documentation for [removing log data with cleanup policies](https://strimzi.io/docs/operators/latest/using.html#removing_log_data_with_cleanup_policies).
```

PaulRMellor: ^^^I've suggested referring to compaction in a note

_posts/2021-12-13-kafka-segment-retention.md
PaulRMellor: I've suggested a compaction note at the start. If you're happy with that, you can remove this.

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,259 @@
...
+The index and timeindex share the same maximum size which is defined by the `log.index.size.max.bytes` configuration parameter and it is 10 MB by default.
+Let's consider the default log segment maximum size, which is 1 GiB.
+We know that because `log.index.interval.bytes` is 4096 bytes by default, an entry is added in the index every 4096 bytes of records.
+It means that for a 1 GiB segment size, 1 GiB / 4096 bytes = 262144 entries are added to the index. This equals 2 MB of index (262144 * 8 bytes).
+The default index size of 10 MB is enough to handle a segment size of 5 GiB.
+This means that reducing the index size or increasing the segment size will mean a new segment is rolled when the index is full and not when the requested segment size is reached.
+So, setting a segment size larger than 5 GiB without increasing the index size is useless, because the broker will roll a new segment when the index is full.
+
+By increasing the segment size over 5 GiB, you would also need to increase the index file size as well. Likewise, if you decide to reduce the index file size, you should probably decrease the segment size as well.
```
mimaison, line 176: Since indexes are 10MB by default, this is extreme optimization. I don't think I'd recommend lowering the index size. Indexes are trimmed once they are closed so the gains of doing so are very minimal (with a potential cost of putting a bad value).

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,259 @@
...
+
+The timeindex might need attention as well. Because each timeindex entry is 1.5x bigger than an entry in the index (12 bytes versus 8 bytes), it can fill up earlier and cause a new segment to be rolled.
```
mimaison, line 178: What do you mean by `need attention`? What should a user do? Especially as both indexes share the same size setting.

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,259 @@
...
+-rw-rw-r--.  1 ppatiern ppatiern 7314 Dec 10 16:23 00000000000000000000.log
+-rw-rw-r--.  1 ppatiern ppatiern  288 Dec 10 16:23 00000000000000000000.timeindex
+-rw-rw-r--.  1 ppatiern ppatiern  296 Dec 10 16:23 00000000000000000049.index
+-rw-rw-r--.  1 ppatiern ppatiern 4500 Dec 10 16:26 00000000000000000049.log
+-rw-rw-r--.  1 ppatiern ppatiern   10 Dec 10 16:23 00000000000000000049.snapshot
+-rw-rw-r--.  1 ppatiern ppatiern  300 Dec 10 16:23 00000000000000000049.timeindex
+```
+
+From the example, a new segment was rolled when the active one was still 7314 byes, not reaching the configured 16384 bytes.
+At the same time, the index reached 192 byes in size, so actually having 192 / 8 = 24 entries and not the expected 37.
```
mimaison, line 197: `byes` -> `bytes`

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,259 @@
...
+
+By increasing the segment size over 5 GiB, you would also need to increase the index file size as well. Likewise, if you decide to reduce the index file size, you should probably decrease the segment size as well.
+
+The timeindex might need attention as well. Because each timeindex entry is 1.5x bigger than an entry in the index (12 bytes versus 8 bytes), it can fill up earlier and cause a new segment to be rolled.
+
+To show how the index and timeindex files size have an impact on rolling a new log segment, let's consider a cluster with `log.index.interval.bytes=150` and `log.index.size.max.bytes=300` still using the Strimzi Canary to produce and consume records with the usual configuration `retention.ms=600000;segment.bytes=16384`.
+Because the Strimzi Canary records are close to 150 bytes in size, we would expect the index file to be filled with one entry every two records.
+With the maximum size set at 300 bytes, we would have around 300 / 8 = 37 entries in the index file and around 300 / 12 = 25 entries in the timeindex file.
```
mimaison, line 182: These last 2 paragraphs are not easy to follow. 
What about something like this:
- say the timeindex is the index that can cause roll because it uses a longer keys than other index as described above
- compute how many entries it can contain with the size set
- extrapolate how many records this maps to
- look at example to confirm roll happened before segment was full due to index filling up first

_posts/2021-12-09-kafka-segment-retention.md
ppatierno: @PaulRMellor but it's not mentioned or at least not clear that this blog post is valid when policy is "delete". I mean, we are referring to "compact" policy in a note with no context of why we should mention it.

_posts/2021-12-09-kafka-segment-retention.md
ppatierno: I agree, but I am not giving any recommendations. I am just highlighting what you should do if you decide to reduce the index file size on your own.

_posts/2021-12-09-kafka-segment-retention.md
ppatierno: It's a poor way of writing what I meant for sure :-)
There is no needed intervention from a user pov. It was a kind of introduction to the next sentence ..."Because each ...." so raising the point that you should not focus just on index file but even on timeindex because it can be filled earlier.
So it needs "attention" :-D

_posts/2021-12-09-kafka-segment-retention.md
ppatierno: > say the timeindex is the index that can cause roll because it uses a longer keys than other index as described above

I think I addressed it on line 178 with "Because each timeindex entry is 1.5x bigger than an entry in the index (12 bytes versus 8 bytes), it can fill up earlier and cause a new segment to be rolled."

> extrapolate how many records this maps to

Not sure what you mean there is no references to the number of records. I am not mentioning them at all, is that important? I said something like "index file to be filled with one entry every two records." ... maybe this information is not useful at all in this context. We don't want to explain in terms of records but just highlighting that the log segment was rolled before 16k because of the index/timeindex full.

_posts/2021-12-09-kafka-segment-retention.md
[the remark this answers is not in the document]
PaulRMellor: Ah, okay. I've rewritten the suggestion.

_posts/2021-12-09-kafka-segment-retention.md
mimaison: To me `you should probably decrease` sounds like a recommendation. @PaulRMellor WDYT?

_posts/2021-12-09-kafka-segment-retention.md
mimaison: No worries, that was just a suggestion.

_posts/2021-12-09-kafka-segment-retention.md
```diff
@@ -0,0 +1,259 @@
...
+
+In our example, it might be that the first record in the segment could live up to almost 24 minutes!
+
+Assuming that at some point the clean thread runs and verifies that a closed segment can be deleted. Initially, it adds a `.deleted` extension to the corresponding files, but doesn't actually delete the segment from the file system.
+The `log.segment.delete.delay.ms` broker parameter defines when the file will actually be removed from the file system when it's marked as "deleted" (1 minute by default).
+
+![Total Retention Time](/assets/images/posts/2021-12-09-total-retention-time.png)
+
+Returning to the Canary example again and assuming the delay on deletion, the first record in our segment is still live after 25 minutes!
+It is really longer than the 10 minutes expectation, isn't it?
```
PaulRMellor, line 241:
```suggestion
It is much longer than the 10 minutes expectation, isn't it?
```

PaulRMellor (approved):
Great :+1:  
I've made a few more suggestion.

_posts/2021-12-09-kafka-segment-retention.md
PaulRMellor: I've adjusted the suggestion.

_posts/2021-12-09-kafka-segment-retention.md
PaulRMellor: @mimaison - It reads like you would probably do it to be optimal. Maybe better to remove it.

PaulRMellor: Otherwise, maybe:  _Likewise, if you decide to reduce the index file size, it is possible that you might want to decrease the segment size proportionately._

ppatierno: @mimaison I would leave the case but using the sentence proposed by Paul instead of removing it. Wdyt?

ppatierno:
Now that I have all approvals, I rescheduled the publishing data as coming Friday (December 17th) as a good reading for the weekend for people :-)
I will just merge over lunch time CEST to wait US friends wake up.

PR #276, opened 2021-12-21
web pages: updates in-development docs to use configuring doc

PaulRMellor:

Updates the doc web pages to use configuring guide instead of using guide for in-development docs.

_Select the type of your PR_

* [ ] Typo/minor fix
* [x] Other


scholzj (commented):
I think you will also need to add HTTP 301 redirect from the old URl to the new URL (in the `_redirects` file).

PaulRMellor:
> I think you will also need to add HTTP 301 redirect from the old URl to the new URL (in the `_redirects` file).

@scholzj - I've added redirects for _in-development_. I guess we update redirects for _latest_ after the next release.

scholzj:
> @scholzj - I've added redirects for in-development. I guess we update redirects for latest after the next release.

Yes, correct.

PR #293, opened 2022-04-15
Add footnote about CRD v1beta1 support in the table with supported versions

scholzj:
The supported versions table in the Download section of the website does not make it really clear which versions require `v1beta1` CRD API. This PR adds a footnote about this and links it from the corresponding versions. This should make the users checking it more aware of this.

Screenshot 2022-04-15 at 17 56 34


dadufour:
Very useful indeed. I would just reword the footnote as "it is not supported anymore in kubernetes as of version 1.22". It is a negative sentence that gets more.
But maybe it is a bit cosmetic ;)

scholzj:
@dadufour Better?

dadufour:
@scholzj : perfect!

ppatierno (approved):

PR #354, opened 2023-03-16
docs(toc): improve usability of documentation ToC

PaulRMellor:
### Type of change

**Strimzi doc updates**
Adds a new ToC to the Strimzi docs that is fixed (follows content) and collapsible based on selection.
Introduces a ToC for mobile sized screens too -- the sticky ToC to the side is replaced by a fixed ToC above the contents.
Plus a couple of cosmetic changes to docs pages

* [ ] Typo/minor fix
* [ ] Other


scholzj (commented):
I do not really use these docs, I use normally only the standalone docs. But some comments:
* It is distracting to me that the ToC on the side is changing all the time when scrolling throgh the docs (this is very subjective)
* The font size does not work for me visually (its weird to have it in much smaller font than the actual documentation text -> I would expect it the same or bigger) (I guess this is subjective as well)
* I liked the original design more (the different background and font sizing) (again, subjective)
* The wrapping does not work. You need to make it fixed with with wrapping I guess. See the screenshot I attached how it overlows out of the screen. (I guess this one is objective ;-))
  Screenshot 2023-03-16 at 16 41 48

ppatierno:
I don't think it is that bad ... maybe the font should be increased a little bit.

PaulRMellor:
> I do not really use these docs, I use normally only the standalone docs. But some comments:
> 
> * It is distracting to me that the ToC on the side is changing all the time when scrolling throgh the docs (this is very subjective)
> * The font size does not work for me visually (its weird to have it in much smaller font than the actual documentation text -> I would expect it the same or bigger) (I guess this is subjective as well)
> * I liked the original design more (the different background and font sizing) (again, subjective)
> * The wrapping does not work. You need to make it fixed with with wrapping I guess. See the screenshot I attached how it overlows out of the screen. (I guess this one is objective ;-))
>   Screenshot 2023-03-16 at 16 41 48

I've changed the TOC approach in the latest commit, which I think addresses these issues.

 - Updated table of contents (TOC) on doc pages
 - Switched from tocbot TOC to fixed TOC similar to asciidoctor
 - Fixed TOC now on left side and scrollable
 - Navigation header also in fixed position
 - Tested changes on Chrome and Firefox at different sizes, no issues found
 - Propose removing asciidoctor versions of TOC from site if this is suitable
 - Note: Footer may need attention in the future (too big, wrong color)

scholzj (commented):
Some comments from my side:
* The indentation of the ToC seems too small. It does not help the readability and just makes it look misaligned. I would use more space for the indentation to make it more clear and maybe make the font smaller so that it still fits.
* Making the ToC items bold when mouse hovers over cause sometimes change of wrapping which cause the whole ToC shift down and back which is confusing. I would not use the bold.
* The header is weird. If I just get tot he documentation page, it looks fine. But the transition from the regular website header to the wide website header in the docs is weirdThe menu is there, but it is still as if you leave the website. But I guess I can live with it if that is fine for the others.
* The footer seems fine, but it should not scroll in over the ToC. 
* I would probably prefer the ToC to be fixed with the content and not have two separate scrollable panes. It seems weird to have essentially one static (header) and three movable panes (ToC, actual docs and footer). But again, if everyone else is fine with it, I can live with it.

PaulRMellor:
> Some comments from my side:
> 
> * The indentation of the ToC seems too small. It does not help the readability and just makes it look misaligned. I would use more space for the indentation to make it more clear and maybe make the font smaller so that it still fits.
> * Making the ToC items bold when mouse hovers over cause sometimes change of wrapping which cause the whole ToC shift down and back which is confusing. I would not use the bold.
> * The header is weird. If I just get tot he documentation page, it looks fine. But the transition from the regular website header to the wide website header in the docs is weirdThe menu is there, but it is still as if you leave the website. But I guess I can live with it if that is fine for the others.
> * The footer seems fine, but it should not scroll in over the ToC.
> * I would probably prefer the ToC to be fixed with the content and not have two separate scrollable panes. It seems weird to have essentially one static (header) and three movable panes (ToC, actual docs and footer). But again, if everyone else is fine with it, I can live with it.

Thanks @scholzj  -- I have updated as follows:

- Increased indentation of ToC
- Removed the bold (much better)
- Added a height constraint on the ToC when the footer is visible when scrolling to the bottom of the page. That is, the footer doesn't go over it. It seems fine at different sizes. So you can always scroll to any part of the ToC now even when the footer is visible.
- I wasn't getting any weird header experience, but I've adjusted so that the TOC reverts to the top whenever the burger menu is visible. Previously, it was on top at a certain width and to the side for a certain width before changing to the mid-width menu.

scholzj (approved):
Some more remarks:
* The Bridge docs has the ToC with horizontal scroll bars because the titles contain too long unbreakable text. Maybe that can be fixed (in a future Bridge release) by not using the full path in the title? If not, maybe we just have to live with it.
* I still don't like ho the site stretches across the whole window on the documentation page. It would be great if we could keep it the same width as on the other pages.

Overall, I can probably live with this if needed. But I don't think it should be merged purely on my approval. It would be good to have opinions and reviews also from others.

ppatierno:
I think this is really a great improvement!

> I still don't like ho the site stretches across the whole window on the documentation page. It would be great if we could keep it the same width as on the other pages.

I think that to have the header not stretching would be enough. For the doc it's much better it takes as much space as possible to be more readable imho.

ppatierno (approved):

Frawless (approved):
LGTM, the overall layout, and ToC look much better from my POV!

im-konge:
Hi @PaulRMellor , I'm looking at the changes and they looks great!
Btw, I'm missing the "search" bar, which is available in the current ToC - is it something that was agreed to remove it? Or is it not used? 
I guess it comes handy in cases we are finding something in the ToC (but yeah, we can use the `Ctrl + F` search tool).

fvaleri (approved):
LGTM. I prefer this to the old layout. Thanks.

PaulRMellor:
Hi @scholzj  -- The latest commit fixes these two issues.

- Titles with long words are now broken apart and wrapped so the horizontal scroll is removed
- The navigation header is now the same as the other pages. I'm not sure the logo looks right where it is, but at least it's the same. I think that's all we should do in terms of stretching. The doc still uses as much space as possible underneath.

Let me know if the title wraps are acceptable

> Some more remarks:
> 
> * The Bridge docs has the ToC with horizontal scroll bars because the titles contain too long unbreakable text. Maybe that can be fixed (in a future Bridge release) by not using the full path in the title? If not, maybe we just have to live with it.
> * I still don't like ho the site stretches across the whole window on the documentation page. It would be great if we could keep it the same width as on the other pages.
> 
> Overall, I can probably live with this if needed. But I don't think it should be merged purely on my approval. It would be good to have opinions and reviews also from others.

PaulRMellor:
Hi @im-konge -- The search bar was not great. It just returned titles in the ToC, which we don't need if we show the full ToC anyway. As you say that can be searched.

> Hi @PaulRMellor , I'm looking at the changes and they look great! Btw, I'm missing the "search" bar, which is available in the current ToC - is it something that was agreed to remove it? Or is it not used? I guess it comes handy in cases we are finding something in the ToC (but yeah, we can use the `Ctrl + F` search tool).

im-konge:
@PaulRMellor thanks for the confirmation and info, that makes sense :)

im-konge (approved):

scholzj (approved):
That looks good @PaulRMellor ... thanks for working through the comments.

PR #383, opened 2023-08-31
Kafka Node Pools: What's next?

scholzj:
This Pr delivered the last blog post for the Node Pools series called _What's next?_. It covers some of the next steps and future improvements that might happen.

_(It will be published only after the fourth part from #382)_

### Type of change

* [x] New blog post (see the [README](https://github.com/strimzi/strimzi.github.io/#blog-posts) for the process)

PR #384, opened 2023-09-06
docs(blog): adds a blog post about developing a producer client

PaulRMellor:
**Blog post**

A post describing a basic producer client
A follow-up post will describe a consumer client

* [ ] Typo/minor fix
* [ ] Other


_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+We can specify the connection details and the name of the topic in our client configuration.
+
+### Client constants
+
+Now, let's define some customizable constants that we'll also use with the producer.
+
+**BOOTSTRAP_SERVERS**
+
+An address and port where the producer client connects to the Kafka cluster is always needed.
```
fvaleri, line 59: This is a list of host/port pairs to use for establishing the initial connection to the Kafka cluster. In production, you can use a single load balancer, or a list of brokers.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+An address and port where the producer client connects to the Kafka cluster is always needed. 
+For example, localhost:9092 might be your starting point.
+We can define it with this constant.
+
+**TOPIC_NAME**
+
+The name of the topic where the producer client sends its messages.
+
+**NUM_MESSAGES**:
```
fvaleri, line 67:
```suggestion
**NUM_MESSAGES**
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+      
+This constant sets the number of messages the client produces before it hits the pause button.
```
fvaleri, line 69:
```suggestion
This constant sets the number of messages the client produces before it shuts down.
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+**MESSAGE_SIZE_BYTES**
+
+We'll use this constant to adjust the size of each message in bytes.
```
fvaleri, line 73:
```suggestion
We'll use this constant to set the size of each message in bytes.
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+These constants give us some control over the producer client's behavior. 
+
+### Example producer client
+
+Time to create our client.
+We want our example client to operate as follows:
+
+* Create a Kafka producer instance, which is responsible for sending messages to a Kafka topic.
+* Generate random message payloads, represented as byte arrays, that serve as the content of the messages being sent to Kafka cluster.
+* Use serializer classes that handle the transformation of message keys (IDs) and values (byte arrays) into a format suitable for the Kafka brokers.
```
fvaleri, line 90: This is a bit confusing. Ultimately, both key and value are serialized into byte arrays.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+* Create a Kafka producer instance, which is responsible for sending messages to a Kafka topic.
+* Generate random message payloads, represented as byte arrays, that serve as the content of the messages being sent to Kafka cluster.
+* Use serializer classes that handle the transformation of message keys (IDs) and values (byte arrays) into a format suitable for the Kafka brokers. 
+* Control the rate at which messages are produced by introducing delays between each message using our `PROCESSING_DELAY_MS` value.
+* Handle errors that may occur during message transmission to the Kafka broker, determining when a message should be retried and when an error is considered non-recoverable. 
+
+**Producer configuration**
+
+We'll specify the following configuration properties for our producer instance:
```
fvaleri, line 96:
```suggestion
At a minimum, you need to specify the following configuration properties for a producer instance:
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+* `KEY_SERIALIZER_CLASS_CONFIG` and `VALUE_SERIALIZER_CLASS_CONFIG` to specify serializers that transform messages into a format suitable for Kafka brokers. 
+In this case, we'll specify the `ByteArraySerializer` as we want to transform byte array values.
+
+We'll also include methods that help with these operations:
+
+**`sleep` method**
+
+- Introduces a delay in the message-sending process for a specified number of milliseconds.
+- Useful for controlling the message production rate or simulating message processing times.
+- Handles potential `InterruptedException` if the thread is interrupted while paused.
```
fvaleri, line 109: The `Thread.sleep` method throws `InterruptedException`, you need to handle them.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+- Determines whether to retry sending a message following a message sending error.
+- Returns `true` if the message sending process can be retried.
+- Returns `false` for null and specified exceptions, or those that do not implement the `RetriableException` interface.
+- Customizable to include other errors.
+
+**`onCompletion` method**
+
+- Confirms successful message transmission and displays information about the message sent, including the topic, partition, and offset.
+- Prints an error message on exception. Appropriate action is taken based on whether it's a fatal or non-fatal error. If the error is non-fatal, the message sending process continues. If the error is fatal, a stack trace is printed and the producer is terminated.
```
fvaleri, line 126:
```suggestion
- Prints an error message on exception. Appropriate action is taken based on whether it's a retriable or non-retriable error. If the error is retriable, the message sending process continues. If the error is non-retriable, a stack trace is printed and the application is terminated.
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+    public void run() {
+        System.out.println("Running producer");
+        try (var producer = createKafkaProducer()) {
+            // Create a Kafka producer instance
+            // This producer sends messages to the Kafka topic asynchronously
+
+            byte[] value = randomBytes(MESSAGE_SIZE_BYTES);
+            while (messageCount.get() < NUM_MESSAGES) {
+                // Generate a random byte array as the message payload
```
fvaleri, line 170: Shouldn't this be on line 168?

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+    protected AtomicLong messageCount = new AtomicLong(0);
+
+    public static void main(String[] args) {
+        new Producer().run();
+    }
+
+    public void run() {
+        System.out.println("Running producer");
+        try (var producer = createKafkaProducer()) {
+            // Create a Kafka producer instance
```
fvaleri, line 165: We can put comments above or below, but we should be consistent on that. My preference is right above.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+        props.put(ProducerConfig.KEY_SERIALIZER_CLASS_CONFIG, LongSerializer.class);
+        props.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG, ByteArraySerializer.class);
+        // Configure serializers for keys and values
+
+        return new KafkaProducer<>(props);
+    }
+
+    private void sleep(long ms) {
+        // sleep method
```
fvaleri, line 200: I think this is useless.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+        try {
+            TimeUnit.MILLISECONDS.sleep(ms);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private byte[] randomBytes(int size) {
+        // randomBytes method
```
fvaleri, line 209: I think this is useless.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+        }
+        byte[] payload = new byte[size];
+        for (int i = 0; i < payload.length; ++i) {
+            payload[i] = (byte) (RND.nextInt(26) + 65);
+        }
+        return payload;
+    }
+
+    private boolean retriable(Exception e) {
+        // retriable method
```
fvaleri, line 221: I think this is useless.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+When running, it creates the message payloads using randomly generated byte arrays. 
+The client produces messages until it reaches the predefined message count, which is 100 messages with the `NUM_MESSAGES` constant value we specified. 
+
+> With its thread-safe design, multiple threads can share a single producer instance.
+
+### Error handling
+
+When developing a Kafka producer client, it's important to consider how you want it to handle different types of exceptions. 
+
+The error handling capabilities we introduced ensures that the producer client can recover from certain non-fatal errors while addressing others as fatal, terminating operation of the client when necessary.
```
fvaleri, line 266:
```suggestion
The error handling capabilities we introduced ensures that the producer client can recover from certain retriable errors while addressing others as non-retriable, terminating operation of the client. 
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+Interruption typically happens during producer shutdown or when stopping its operation. 
+The exception is rethrown as a `RuntimeException`, which ultimately terminates the producer.
+
+* `IllegalArgumentException`: This error is thrown when the producer receives invalid or inappropriate arguments. 
+For instance, it can be triggered if essential details like the topic are missing.
+
+* `UnsupportedOperationException`: This error is raised when an operation is not supported or when a method is not implemented as expected. 
+For instance, it can be triggered if you attempt to use an unsupported producer configuration or invoke a method that the `KafkaProducer` class does not support.
+
+**Non-fatal errors caught by the producer client**
```
fvaleri, line 282:
```suggestion
**Retriable errors caught by the producer client**
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+The example Kafka producer client in this post serves as a foundation. 
+Feel free to build on it.
+For instance, you might want to add custom functionality for integration with your preferred logging framework.
+
+You might also want to explore how to expand and improve on other aspects of your client through configuration:
+
+**Implementing security**
+
+Implement security to establish a secure connection using authentication and authorization mechanisms when connecting to the Kafka cluster. 
+For example, you can set up TLS authentication for external clients in your Strimzi environment and add the TLS certificates to your client configuration.
```
fvaleri, line 297: Here I would mention an hidden Strimzi gem: https://github.com/strimzi/kafka-kubernetes-config-provider.

fvaleri (commented):
Hi @PaulRMellor, thanks. I left some comments.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+
+Let's break down the essential steps to get started:
+
+1. Start by choosing the Kafka client library that speaks your choice of programming language. We use Java in this post, but you can use Python, .NET, and so on.
+
+2. Get the library through a package manager or by downloading it from the source.
+
+3. Import the classes and dependencies that your Kafka client will need into your code.
+
+4. Tell your client how to find and connect with your Kafka cluster, specifying configuration properties for broker address and port, and, if required, security credentials.
```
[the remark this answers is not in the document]
PaulRMellor: This is more generally about clients and we haven't spoken about the bootstrap server property yet. Maybe we can just say " specifying an address and port"  instead

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,324 @@
...
+Let's get down to creating the producer client.
+Our brief is to create a client that operates asynchronously, and is equipped with basic error-handling capabilities. 
+To handle messages, it makes use of Kafka's `Callback` interface. 
+
+### Prerequisites
+
+To be able to operate, the producer client needs the following in place:
+
+* Running kafka brokers 
+* A Kafka topic where it sends messages
```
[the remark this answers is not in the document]
PaulRMellor: we can include this in a note. Either way, the client does need a topic.
Note:

> By default, cluster configuration has `auto.create.topics.enable=true`, which means a topic is automatically created on the first request.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I took the bracketed bits out

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: all on top now

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -296,11 +291,14 @@ You might also want to explore how to expand and improve on other aspects of you
 Implement security to establish a secure connection using authentication and authorization mechanisms when connecting to the Kafka cluster. 
 For example, you can set up TLS authentication for external clients in your Strimzi environment and add the TLS certificates to your client configuration. 
 
+You can use configuration providers to load configuration, including secrets, from external sources.
```
fvaleri, line 294: As an example, I would mention specifically the `KubernetesSecretConfigProvider` to extract TLS certificates and keys directly from the Kubernetes API. This is handy, because you won't need to do any configuration change, just a simple application restart (hot reloading is not supported right now).

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,322 @@
...
+
+2. Get the library through a package manager or by downloading it from the source.
+
+3. Import the classes and dependencies that your Kafka client will need into your code.
+
+4. Tell your client how to find and connect with your Kafka cluster, specifying an address and port, and, if required, security credentials.
+
+5. Create a producer object to subscribe to topics and produce messages to Kafka.
+
+    > A client can be a Kafka producer, and consumer, and admin.
```
scholzj, line 34: I would maybe add Streams API / stream processing.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Client constants
+
+Now, let's define some customizable constants that we'll also use with the producer.
+
+**BOOTSTRAP_SERVERS**
+
+With this constant, we can define a list of host/port pairs for establishing an initial connection to the Kafka cluster.
+For example, `localhost:9092` might be your starting point.
+In production, you can use a single load balancer, or a list of brokers.
```
scholzj, line 63: I have a bit of mixed feelings about this. I guess using `localhost:9092` makes things easier. But it also means it has no real relationship to Strimzi and it does not help users to connect this to Strimzi. So it is basically just a generic _Kafka producer_ guide.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,322 @@
...
+* Control the rate at which messages are produced by introducing delays between each message using our `PROCESSING_DELAY_MS` value.
+* Handle errors that may occur during message transmission to the Kafka broker, determining when a message should be retried and when an error is considered non-recoverable. 
+
+**Producer configuration**
+
+We'll specify the minimal configuration properties required for a producer instance:
+
+* `BOOTSTRAP_SERVERS_CONFIG` for connection to the Kafka brokers. This picks up the value of the `BOOTSTRAP_SERVERS` constant.
+* `CLIENT_ID_CONFIG` that uses a randomly generated UUID as a client ID for tracking the source of requests.
+* `KEY_SERIALIZER_CLASS_CONFIG` and `VALUE_SERIALIZER_CLASS_CONFIG` to specify serializers that transform messages into a format suitable for Kafka brokers.
```
scholzj, line 102: People not familiar with the Kafka clients might not be familiar with this either.  So I wonder if it would make sense to dive into more detail here. I would for example mention examples of other available (de)serializers, possibility to write custom ones for more complicated structures etc.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,322 @@
...
+
+## Tuning your producer
+
+The example Kafka producer client in this post serves as a foundation. 
+Feel free to build on it.
+For instance, you might want to add custom functionality for integration with your preferred logging framework.
+
+You might also want to explore how to expand and improve on other aspects of your client through configuration:
+
+**Implementing security**
```
scholzj, line 289: Again, I wonder if Strimzi should be plugged into this. Configuring the security with Strimzi is a pretty common struggle, but you basically just jump over it.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,322 @@
...
+**Improving data durability** 
+  
+Specify `acks=all` (default) in your producer configuration so that all in-sync topic replicas acknowledge successful message delivery. 
+Or configure `transaction` properties in your brokers and producer client application to ensure that messages are processed in a single transaction.
+
+**Boosting performance** 
+  
+Optimize your producer for high message throughput and low latency. 
+Use the `linger.ms` and `batch.size` configuration properties to batch more messages into a single produce request for higher throughput. 
+Improve throughput of your message requests by using the `delivery.timeout.ms` property to adjust the maximum time to wait before a message is delivered and completes a send request.
```
scholzj, line 308: Didn't we had a blog post about performance tuning? Maybe you can link to it?

scholzj (commented):
Some general notes:
* I would at least touch on the dependencies that need to be added to the Java project (i.e. the Kafka client JAR).
* Would it make sense to mention at the start that this expects some basic Java knowledge / experience?

scholzj:
I wonder if @mimaison, @tombentley, or @showuon would be interested in looking into this as well. It is basically more Kafka than Strimzi, so I wonder if they know of some common misunderstanding among Kafka users which would be worth explaining etc.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I extended the description.

> **BOOTSTRAP_SERVERS**
> 
> With this constant, we define the initial connection point to the Kafka cluster. 
> You can specify a list of host/port pairs to establish this connection.
> For a generic Kafka deployment, you might start with a value like `localhost:9092`. 
> However, when working with a Strimzi-managed Kafka cluster, it's important to use the bootstrap server address configured in your Strimzi deployment.
> 
> In Strimzi, the bootstrap server address is in the format `<cluster_name>-kafka-<listener-name>-bootstrap:<port>`. 
> This address allows an external Kafka producer to connect to the cluster through an advertised listener port (9092 or higher).
> In production environments, you can use a single load balancer, or a list of brokers in place of a single bootstrap server.

scholzj: Maybe you should point them instead to get the bootstrap address from the Kafka CR status? As that would handle the external listeners as well, which are needed if you don't include how to build this into Docker image and deploy in Kubernetes

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+date: 2023-09-15
+author: paul_mellor
+---
+
+Strimzi simplifies the deployment and management of Kafka clusters in a Kubernetes environment, ensuring a smooth and hassle-free experience.
+But the real fun begins once your cluster is up and running.
+That's when you can start thinking about how to interact with Kafka to stream data.
+In other words, Kafka doing the job it excels at.
+
+In this post, we'll dive into the essentials of developing a Kafka producer client that can send messages to a Strimzi-managed Kafka cluster.
```
mimaison, line 13: It look like this post is about developing a Kafka producer client _application_, or just Kafka producer application.
For me a "Kafka producer client" is the underlying client that speaks the Kafka protocol, it's an interface between systems. On the other hand, an application achieves something, a use case , a process, for its users. If you agree, then the title should be updated as well and pretty much replace all instances of "client" by "application".

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Getting started with your producer client
+
+The first thing to consider when developing a producer client application is your preferred programming language. 
+
+After you've decided on that, as a bare minimum, your application must be able to connect to the Kafka cluster and use producers to send messages.
+
+Let's break down the essential steps to get started:
+
+1. Start by choosing the Kafka client library that speaks your choice of programming language. We use Java in this post, but you can use Python, .NET, and so on.
```
mimaison, line 24: This may be obvious but I wonder if it's worth mentioning the the Java client is the official client that is part of the Apache Kafka project. All other clients are third party and may not have the same features, compatibility guarantees and behavior.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+2. Get the library through a package manager or by downloading it from the source.
+
+3. Import the classes and dependencies that your Kafka client will need into your code.
+
+4. Tell your client how to find and connect with your Kafka cluster, specifying an address and port, and, if required, security credentials.
+
+5. Create a producer object to subscribe to topics and produce messages to Kafka.
```
mimaison, line 32: Producers don't "subscribe" to topics. Also instance may be a better word than "object".

_posts/2023-09-15-kafka-producer-client-essentials.md
mimaison: Should we use "or" instead of "and". A client is usually one of them, while an application often combines multiple clients.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+    > A client can be a Kafka producer, and consumer, and admin. 
+
+6. Pay attention to error handling. It's tricky, but vitally important when connecting and communicating with Kafka.
```
mimaison, line 36: I wouldn't necessarily say it's "tricky", but error handling is important for production systems where high availability and ease of operations are valued. That's the difference between a prototype and a production grade application and not something specific to Kafka. If the point of the post is to cover error handling/best practices in particular I'd highlight the "for production" aspect in the title or early paragraphs.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+## Creating the Kafka producer client
+
+Let's get down to creating the producer client.
+Our brief is to create a client that operates asynchronously, and is equipped with basic error-handling capabilities. 
+The application implements the `Callback` interface, which provides a method for asynchronous handling of request completion in a background thread.
```
mimaison, line 42: Should we link to the specific `Callback` interface you refer to here? Many framework have `Callback` interface including the Java SDK itself!

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+**MESSAGE_SIZE_BYTES**
+
+We'll use this constant to set the size of each message in bytes.
+
+**PROCESSING_DELAY_MS**
+
+Sometimes, it's good to slow things down a bit. 
+We can use this constant to add a delay in milliseconds between sending messages. 
+Adding a  delay can be useful when testing in order to simulate typical message processing time.
```
mimaison, line 86: It's not really "processing time" but rather "emission/creation time". In Kafka records are typically events that are created as "things happen", so this can help simulate peak or average event rates.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+For example, `localhost:9092` might be your starting point.
+In production, you can use a single load balancer, or a list of brokers.
+
+**TOPIC_NAME**
+
+The name of the topic where the producer client sends its messages.
+
+**NUM_MESSAGES**
+      
+This constant sets the number of messages the client produces before it shuts down.
```
mimaison, line 76: What about" `The number of messages the client produces before it shuts down`?
It's like the description just above. IMO adding "this constant" in these descriptions just makes them harder to read. Also since those are values you can change, they are not really _constants_!

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+- Introduces a delay in the message-sending process for a specified number of milliseconds.
+- Useful for controlling the message production rate or simulating message processing times.
+- Handles potential `InterruptedException` if the thread responsible for sending messages is interrupted while paused.
+
+**`randomBytes` method**
+
+- Generates a random byte array of a specified size to serve as the payload for each message sent to the Kafka topic. Adds 65 to represent an uppercase letter in ASCII code (65 is 'A', 66 is 'B', and so on).
+- Ensures that the payload size is greater than zero and throws an `IllegalArgumentException` if not.
+
+**`retriable` method**
```
mimaison, line 128: Should we explain the impact of retrying? Can this cause duplicates? How does that differ from the producer configurations like `retries` or idempotency?

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+        try {
+            TimeUnit.MILLISECONDS.sleep(ms);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private byte[] randomBytes(int size) {
+        if (size <= 0) {
+            throw new IllegalArgumentException("Record size must be greater than zero");
```
mimaison, line 218: Should we mention the configuration, `MESSAGE_SIZE_BYTES` that has a bad value?

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+        }
+        byte[] payload = new byte[size];
+        for (int i = 0; i < payload.length; ++i) {
+            payload[i] = (byte) (RND.nextInt(26) + 65);
+        }
+        return payload;
+    }
+
+    private boolean retriable(Exception e) {
+        if (e == null) {
```
mimaison, line 228: Is this needed? It seems the caller already checked that the Exception was not null.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+    public void onCompletion(RecordMetadata metadata, Exception e) {
+        if (e != null) {
+            // If an exception occurred while sending the record
+            System.err.println(e.getMessage());
+
+            if (!retriable(e)) {
+                // If the exception is not retriable, print the stack trace and exit
+                e.printStackTrace();
+                System.exit(1);
+            }
```
mimaison, line 249: Should we retry sending this message if the error is retriable?

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+}
+```
+
+### Running the producer client
+
+To put this client into action, simply run the main method in the `Producer` class. 
+When running, it creates the message payloads using randomly generated byte arrays. 
+The client produces messages until it reaches the predefined message count, which is 100 messages with the `NUM_MESSAGES` constant value we specified. 
+
+> With its thread-safe design, multiple threads can share a single producer instance.
```
mimaison, line 265: Maybe link to docs that attest of this statement?

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I've expanded this section:

> **Implementing security**
> 
> Ensure a secure connection when connecting to your Kafka cluster by implementing security measures for authentication, encryption, and authorization. In Strimzi, this process involves configuring listeners and user accounts:
> 
> * Listener configuration: Use the `Kafka` resource to configure listeners for client connections to Kafka brokers.
> Listeners define how clients authenticate, such as using TLS, SCRAM-SHA-512, OAuth 2.0, or custom authentication methods.
> To enhance security, configure TLS encryption to secure communication between Kafka brokers and clients.
> You can further secure TLS-based communication by specifying the supported TLS versions and cipher suites in the Kafka broker configuration.
> For an added layer of protection, you can also specify authorization methods in your listener configuration, such as simple, OAuth 2.0, OPA, or custom authorization.
> 
> * User Accounts: Set up user accounts and credentials with `KafkaUser` resources in Strimzi. 
> Users represent your clients and determine how they should authenticate and authorize with the Kafka cluster. 
> The authentication and authorization mechanisms used must match the Kafka configuration. 
> Additionally, define Access Control Lists (ACLs) to control user access to specific topics and actions for more fine-grained authorization.
> You can also add configuration to limit the TLS versions and cipher suites your client uses.
> 
> Implementing security can be a complex topic, but Strimzi simplifies the process for you. For more information on securing access to Kafka brokers using `Kafka` and `KafkaUser` resources, see the [Strimzi documentation describing how to secure access to Kafka brokers](https://strimzi.io/docs/operators/latest/deploying#assembly-securing-kafka-str).

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+* `InterruptedException`: This error occurs when the current thread is interrupted while paused. 
+Interruption typically happens during producer shutdown or when stopping its operation. 
+The exception is rethrown as a `RuntimeException`, which ultimately terminates the producer.
+
+* `IllegalArgumentException`: This error is thrown when the producer receives invalid or inappropriate arguments. 
+For instance, it can be triggered if essential details like the topic are missing.
+
+* `UnsupportedOperationException`: This error is raised when an operation is not supported or when a method is not implemented as expected. 
+For instance, it can be triggered if you attempt to use an unsupported producer configuration or invoke a method that the `KafkaProducer` class does not support.
```
mimaison, line 285: How can you "invoke a method that the `KafkaProducer` class does not support"? Surely that code would not compile.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I linked to the tuning producers doc.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+**Retriable errors caught by the producer client**
+
+`RetriableException`: This type of error is thrown for any exception that implements the `RetriableException` interface, as provided by the Kafka client library.
+
+## Tuning your producer
+
+The example Kafka producer client in this post serves as a foundation. 
+Feel free to build on it.
+For instance, you might want to add custom functionality for integration with your preferred logging framework.
```
mimaison, line 295: The main points to look on the producer side are:
- compression
- batching
- partitioning
There really impact the performance and behavior and need to be carefully considered when building an application.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+You can use configuration providers to load configuration, including secrets, from external sources.
+For example, you can use the `KubernetesSecretConfigProvider` to extract TLS certificates and keys directly from the Kubernetes API. 
+For more information, see the [Strimzi documentation on configuration providers](https://strimzi.io/docs/operators/latest/deploying#assembly-loading-config-with-providers-str). 
+
+> Configure the security protocol used by your client application to match the protocol configured on a Kafka broker listener. 
+
+**Improving data durability** 
+  
+Specify `acks=all` (default) in your producer configuration so that all in-sync topic replicas acknowledge successful message delivery. 
+Or configure `transaction` properties in your brokers and producer client application to ensure that messages are processed in a single transaction.
```
mimaison, line 324: There's no transaction properties to set on the broker side.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+**Boosting performance**
```
mimaison, line 326: Don't forget compression!

PaulRMellor:
> Some general notes:
> 
> * I would at least touch on the dependencies that need to be added to the Java project (i.e. the Kafka client JAR).
> * Would it make sense to mention at the start that this expects some basic Java knowledge / experience?

Yes. I'll look at these. The second commit ([a698e95](https://github.com/strimzi/strimzi.github.io/pull/384/commits/a698e95a76b1d26cf66ce97a158464a6fbd10d6c)) just went through the initial comments.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+
+**Introducing further error handling** 
+  
+Introduce more fine-grained error handling capabilities that also improve the resilience of the client.
+
+## Send a message
+
+In this blog post, we've explored how to develop a Kafka producer client.
+We've covered the essential steps, illustrated with an example that highlights asynchronous message production and effective error handling.
+Remember, the key to a successful producer client lies in its ability to connect and communicate with a Kafka cluster.
```
mimaison, line 342: Sure connectivity is necessary but that's true for every type of client. For a producer, batching, partitioning are the key aspects.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,323 @@
...
+  
+Introduce more fine-grained error handling capabilities that also improve the resilience of the client.
+
+## Send a message
+
+In this blog post, we've explored how to develop a Kafka producer client.
+We've covered the essential steps, illustrated with an example that highlights asynchronous message production and effective error handling.
+Remember, the key to a successful producer client lies in its ability to connect and communicate with a Kafka cluster. 
+Once you have that foundation, you define how you want your client to produce and send messages.
+For example, Kafka producer clients typically pull data from external sources.
```
mimaison, line 344: I find this statement a bit strange

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Changed to _producer application_.
I think it's better for SEO too, as it appears to be a more commonly used search item.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I'll mention this. The post isn't about error handling in particular. we're just showing an example application.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: changed as follows:

Adding a  delay can be useful when testing in order to simulate typical message creation patterns.
In Kafka, messages typically capture streams of events, so introducing delays can help simulate peak or average event rates.

_posts/2023-09-15-kafka-producer-client-essentials.md
fvaleri: The Kafka producer automatically handles retries for certain errors (i.e. connection errors), and you can tune it with configuration. That said, the application developer also needs to implement some retry logic for business level exceptions.

Yes, Kafka has at-least-once semantics by default, so the only way to avoid the risk of duplicates is to enable transactions. You can link our blog post for more details: https://strimzi.io/blog/2023/05/03/kafka-transactions

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: @fvaleri -- is the check needed?

```
if (e == null) {
   return false;
```

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: @fvaleri - thoughts? We could expand the method or just add a comment about expanding it.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Not sure what docs to link to, so changed the note!

> Kafka producer instances are designed to be thread-safe, allowing multiple threads to share a single producer instances.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Changed to:

> For instance, it can be triggered if you attempt to use an unsupported producer configuration, leading to a runtime exception.

_posts/2023-09-15-kafka-producer-client-essentials.md
fvaleri: @mimaison is right, it is redundant in this case.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I've added something on this.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: changed to:

> Remember, while connectivity is fundamental for client applications, the key to a successful producer client lies in its ability to effectively send messages using mechanisms like batching and by making use of efficient partitioning strategies within the Kafka cluster.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Yeah. It's a bit odd. Removed!

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,372 @@
...
+
+        // Create a Kafka producer instance
+        // This producer sends messages to the Kafka topic asynchronously
+        try (var producer = createKafkaProducer()) {
+            // Generate a random byte array as the message payload
+            byte[] value = randomBytes(MESSAGE_SIZE_BYTES);
+            while (messageCount.get() < NUM_MESSAGES) {
+                sleep(PROCESSING_DELAY_MS);                
+                // Send a message to the Kafka topic, specifying topic name, message count, and message value
+                producer.send(new ProducerRecord<>(TOPIC_NAME, messageCount.get(), value), this);
```
showuon, line 206: This is a little confusing. It looks like the `ProducerRecord` takes a `messageCount` parameter, but we're actually taking `messageCount` as the message key. Could we explain it here?

_posts/2023-09-15-kafka-producer-client-essentials.md
scholzj: > It look like this post is about developing a Kafka producer client _application_, or just Kafka producer application. For me a "Kafka producer client" is the underlying client that speaks the Kafka protocol, it's an interface between systems. On the other hand, an application achieves something, a use case , a process, for its users. If you agree, then the title should be updated as well and pretty much replace all instances of "client" by "application".

@mimaison @PaulRMellor would this change also make sense in the title? `Developing Kafka clients: A simple producer application` => `Developing Kafka client applications: A simple producer`

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,372 @@
...
+
+Let's break down the essential steps to get started:
+
+1. Start by choosing the Kafka client library that speaks your choice of programming language. We use Java in this post, but you can use Python, .NET, and so on. The Java client is part of the Apache Kafka project. 
+
+2. Get the library through a package manager or by downloading it from the source.
+
+3. Import the classes and dependencies that your Kafka client will need into your code.
+
+4. Tell your client how to find and connect with your Kafka cluster, specifying an address and port, and, if required, security credentials.
```
ppatierno, line 33: `specifying an address and port` ... I would say the bootstrap servers list, because in general it's a list of address:port not just one. With Strimzi it's just a service so in the end you will have just one but here we are not mentioning Strimzi yet.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,372 @@
...
+
+Let's get down to creating the producer application.
+Our brief is to create a client that operates asynchronously, and is equipped with basic error-handling capabilities. 
+The application implements the [producer `Callback` interface](https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/producer/Callback.java), which provides a method for asynchronous handling of request completion in a background thread.
+
+### Adding dependencies
+
+Before implementing the Kafka producer application, our project must include the necessary dependencies.
+For a Java-based Kafka client, we need to include the Kafka client JAR. 
+This JAR file contains the Kafka libraries required for building and running the client.
```
ppatierno, line 52: Could that be useful to provide a code snipped with the XML part of the pom.xml file bringing the Kafka client dependency?

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,372 @@
...
+
+### Client constants
+
+Now, let's define some customizable constants that we'll also use with the producer.
+
+**BOOTSTRAP_SERVERS**
+
+The initial connection point to the Kafka cluster. 
+You can specify a list of host/port pairs to establish this connection.
+For a generic Kafka deployment, you might start with a value like `localhost:9092`.
```
ppatierno, line 73: More than "generic" I would say a "local" because using localhost:9092 assumes your client is running locally as well as your Kafka cluster.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,372 @@
...
+
+**MESSAGE_SIZE_BYTES**
+
+The size of each message in bytes.
+
+**PROCESSING_DELAY_MS**
+
+Sometimes, it's good to slow things down a bit. 
+We can use this constant to add a delay in milliseconds between sending messages. 
+Adding a  delay can be useful when testing in order to simulate typical message creation patterns.
```
ppatierno, line 100:
```suggestion
Adding a delay can be useful when testing in order to simulate typical message creation patterns.
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,372 @@
...
+* Handle errors that may occur during message transmission to the Kafka broker, determining when a message should be retried and when an error is considered non-recoverable. 
+
+**Producer configuration**
+
+We'll specify the minimal configuration properties required for a producer instance:
+
+* `BOOTSTRAP_SERVERS_CONFIG` for connection to the Kafka brokers. This picks up the value of the `BOOTSTRAP_SERVERS` constant.
+* `CLIENT_ID_CONFIG` that uses a randomly generated UUID as a client ID for tracking the source of requests.
+* `KEY_SERIALIZER_CLASS_CONFIG` and `VALUE_SERIALIZER_CLASS_CONFIG` to specify serializers that transform messages into a format suitable for Kafka brokers. 
+In this case, we'll specify the `ByteArraySerializer` as we want to transform byte array values.
```
ppatierno, line 123: `we want to transform byte array values.` ... not sure that "transform" is valid here. 
I mean, the serializer transforms your data/object/value in bytes to be sent. The `ByteArraySerializer` actually does nothing ... it gets bytes and returns the same bytes to be sent on the wire. Actually it's not transforming "something" into bytes as a more complex serializer would do.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,372 @@
...
+import org.apache.kafka.common.serialization.LongSerializer;
+
+public class Producer implements Callback {
+    // Constants for configuration
+    private static final Random RND = new Random(0);
+    private static final String BOOTSTRAP_SERVERS = "localhost:9092";
+    private static final String TOPIC_NAME = "my-topic";
+    private static final long NUM_MESSAGES = 100;
+    private static final int MESSAGE_SIZE_BYTES = 100;
+    private static final long PROCESSING_DELAY_MS = 0L;
```
ppatierno, line 187: Is it intentional using 0 delay? With this value the producer would work as expected but I would like to make a broader point ...

The `producer.send` is asynchronous and it just puts the data in the buffer to make a batch. It will be sent based on  `batch.size` (default 16384) and `linger.ms` (default 0). Due to default 0 `linger.ms` each batch with one message (just 100 bytes) will be sent immediately.
If the reader changes `PROCESSING_DELAY_MS` with an higher value, i.e. 1000 ms, the code will still work as expected and sending each message every 1000 ms.
Things become complicated if the reader starts playing with above `batch.size` and `linger.ms`.
For example, setting `PROCESSING_DELAY_MS = 100` but `linger.ms = 500` will have the result of not sending each message every 100 ms. But each iteration in the loop would give time to the internal producer buffer to be filled with 4-5 messages (up to the 500ms) and then all sent together.
In conclusion, we should not make the blog post much complicated but at least one note for the user to be aware of how `batch.size` and `linger.ms` can have an influence on the expected `PROCESSING_DELAY_MS`?

Also, when running the example the client will send 100 messages in a really short time. Maybe we could have less messages to send and an higher delay between each message to give time to the user seeing the producer sending messages (i.e. 10 or 50 messages every 1000 ms?).

_posts/2023-09-15-kafka-producer-client-essentials.md
mimaison: Good suggestion, I agree

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Thanks @fvaleri -- I changed the note as follows:

> By default, Kafka operates with at-least-once delivery semantics, which means that messages can be delivered more than once in certain scenarios, potentially leading to duplicates. 
> To mitigate this risk, consider enabling transactions in your Kafka producer. 
> Transactions provide stronger guarantees of exactly-once delivery. 
> Additionally, you can use the `retries` configuration property to control how many times the producer will retry sending a message before giving up. 
> This setting affects how many times the `retriable` method may return true during a message send error. 
> For more information, see the [Strimzi post on transactions](https://strimzi.io/blog/2023/05/03/kafka-transactions/) and the [Strimzi documentation on ordered delivery](https://strimzi.io/docs/operators/latest/deploying#ordered_delivery).

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: @fvaleri - Thanks. Changed as follows:

```
private boolean retriable(Exception e) {
        if (e instanceof IllegalArgumentException
            || e instanceof UnsupportedOperationException
            || !(e instanceof RetriableException)) {
            return false;
        } else {
            return true;
        }
    }
```

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Thanks @fvaleri . I updated the description of the `retriable` method

> **`retriable` method**
> 
> - Determines whether to retry sending a message following a message sending error.
> - Returns `true` if the message sending process can be retried. The Kafka producer automatically handles retries for certain errors, such as connection errors.
> - Returns `false` for null and specified exceptions, or those that do not implement the `RetriableException` interface.
> - Customizable to include other errors and  implementing retry logic for business level exceptions.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Changed to:  `// Send a message to the Kafka topic, specifying topic name, message key, and message value`

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I would say so. Added:

To do this in a Java-based project using Apache Maven, add the Kafka client dependency to your `pom.xml` file as follows:

```xml
<dependencies>
    <dependency>
        <groupId>org.apache.kafka</groupId>
        <artifactId>kafka-clients</artifactId>
        <version><kafka_version></version>
    </dependency>
</dependencies>
```

ppatierno: replace `<kafka_version>` with `${kafka.version}`

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: changed as follows:

> In this case, we'll specify the `ByteArraySerializer` because it simply represents the data as bytes without further transformation

ppatierno: Fine with me.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: I've changed the defaults to sending 50 messages every 1000 ms and added a bit more context. 
We have a section on tuning later which discusses linger and batch size.

PaulRMellor: FYI @fvaleri^

ppatierno: yeah but you just mention `Use the `linger.ms` and `batch.size` configuration properties to batch more messages into a single produce request for higher throughput. ` ... there is no explanation about what I said here.

_posts/2023-09-15-kafka-producer-client-essentials.md
mimaison: In case you want a source to reference, the [KafkaProducer Javadoc](https://kafka.apache.org/35/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html) states `The producer is thread safe and sharing a single producer instance across threads will generally be faster than having multiple instances. `

_posts/2023-09-15-kafka-producer-client-essentials.md
mimaison: I wonder if instead you get this is the broker side does not support a feature, for example idempotency. @fvaleri do you remember what scenarios cause `UnsupportedOperationException`?

_posts/2023-09-15-kafka-producer-client-essentials.md
fvaleri: Fine, but I would remove the phrase about `retriable` method, which is unrelated (it comes after all retries are exhausted or timeout).

_posts/2023-09-15-kafka-producer-client-essentials.md
fvaleri: The simple `retriable` method here is used to know when the error is fatal and we need to exit. Otherwise we simply print the error and continue, but we do not retry. This is to keep the example simple, but we can add a note above `if (!retriable(e)) {` line saying that the user should add retry logic for business errors.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: Thanks @ppatierno. I've tried to keep it minimal, as I don't want the post to get so dense it's not easily digestible.

How's this:

The `producer.send` method is asynchronous and buffers messages for batching.
Use the `linger.ms` and `batch.size` configuration properties to batch more messages into a single produce request for higher throughput. 
Modifying these properties impacts the message sending rate in relation to the `PROCESSING_DELAY_MS` setting.
For example, if you set `PROCESSING_DELAY_MS` to 1000 ms and `linger.ms` to 5000 ms, messages are batched and sent every 5000 ms (as specified by `linger.ms`).

ppatierno (approved):
Nice work! LGTM!

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,394 @@
...
+This JAR file contains the Kafka libraries required for building and running the client.
+
+To do this in a Java-based project using Apache Maven, add the Kafka client dependency to your `pom.xml` file as follows:
+
+```xml
+<dependencies>
+    <dependency>
+        <groupId>org.apache.kafka</groupId>
+        <artifactId>kafka-clients</artifactId>
+        <version><kafka_version></version>
```
fvaleri, line 61: Why not use the current latest Kafka release here, so that the early readers can simply cut and paste this configuration without change? The example may not work with a future Kafka release.

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,394 @@
...
+    </dependency>
+</dependencies>
+```
+### Prerequisites
+
+To be able to operate, the producer application needs the following in place:
+
+* A running kafka cluster
```
fvaleri, line 69:
```suggestion
* A running Kafka cluster 
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,394 @@
...
+We can use the `NUM_MESSAGES` and `PROCESSING_DELAY_MS` values to adjust the message sending rate.
+With the current settings, at total of 50 messages are sent with a delay of 1 second between them.
+
+### Example producer application
+
+Time to create our client.
+We want our example client to operate as follows:
+
+* Create a Kafka producer instance, which is responsible for sending messages to a Kafka topic.
+* Generate random message payloads, represented as byte arrays, that serve as the content of the messages being sent to Kafka cluster.
```
fvaleri, line 124:
```suggestion
* Generate a random message payload, represented as byte arrays, that serve as the content of the messages being sent to Kafka cluster.
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,394 @@
...
+
+**`retriable` method**
+
+- Determines whether to retry sending a message following a message sending error.
+- Returns `true` if the message sending process can be retried. The Kafka producer automatically handles retries for certain errors, such as connection errors.
+- Returns `false` for null and specified exceptions, or those that do not implement the `RetriableException` interface.
+- Customizable to include other errors and  implementing retry logic for business level exceptions.
+
+> By default, Kafka operates with at-least-once delivery semantics, which means that messages can be delivered more than once in certain scenarios, potentially leading to duplicates. 
+> To mitigate this risk, consider enabling transactions in your Kafka producer.
```
fvaleri, line 164:
```suggestion
> To avoid this risk, consider enabling transactions in your Kafka producer. 
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,394 @@
...
+                    metadata.topic(), metadata.partition(), metadata.offset());
+        }
+    }
+}
+```
+
+### Running the producer application
+
+To put this client into action, simply run the main method in the `Producer` class. 
+When running, it creates the message payloads using randomly generated byte arrays.
```
fvaleri, line 296:
```suggestion
When running, it creates the message payload using a randomly generated byte array. 
```

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,394 @@
...
+
+You can also improve throughput of your message requests by using the `delivery.timeout.ms` property to adjust the maximum time to wait before a message is delivered and completes a send request. 
+
+For more information, see the [Strimzi documentation on tuning producers](https://strimzi.io/docs/operators/latest/deploying#con-producer-config-properties-str).
+
+**Introducing further error handling** 
+  
+Introduce more fine-grained error handling capabilities that also improve the resilience of the client.
+
+## Send a message
```
fvaleri, line 388: I guess this should be "Conclusion" or something like that.

fvaleri (commented):
Thanks @PaulRMellor, great work!

Last comments from me.

showuon (approved):
LGTM!

_posts/2023-09-15-kafka-producer-client-essentials.md
mimaison: @PaulRMellor Do you want to update the title to Jakub's suggestion?

_posts/2023-09-15-kafka-producer-client-essentials.md
```diff
@@ -0,0 +1,394 @@
+---
+layout: post
+title: "Developing Kafka clients: A simple producer application"
```
scholzj, line 3: Should we call it `Developing Kafka client applications: A simple producer` or something like that as discussed in https://github.com/strimzi/strimzi.github.io/pull/384#discussion_r1319713209

scholzj (approved):
I left some nits, but LGTM otherwise. I will publish https://github.com/strimzi/strimzi.github.io/pull/383 on Monday. So we should be bale to publish this at the end of the week or something like that.

_posts/2023-09-15-kafka-producer-client-essentials.md
PaulRMellor: It's kind of a call to action. I'm not sure we need to say "Conclusion".

fvaleri (approved):
LGTM. Thanks @PaulRMellor.

_posts/2023-09-15-kafka-producer-client-essentials.md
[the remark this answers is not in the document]
PaulRMellor: Yeah. I just used the 3.5.1

PaulRMellor:
> I left some nits, but LGTM otherwise. I will publish #383 on Monday. So we should be bale to publish this at the end of the week or something like that.

I changed the date to 25/09

scholzj:
> > I left some nits, but LGTM otherwise. I will publish #383 on Monday. So we should be bale to publish this at the end of the week or something like that.
> 
> I changed the date to 25/09

You should probably rename the file as well.

mimaison (approved):

PR #395, opened 2023-10-23
Add Unidirectional Topic Operator blog post

fvaleri:
The blog post also contains some load testing stats and instructions to upgrade/downgrade.


_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+title: "Unidirectional Topic Operator"
+date: 2023-10-23
+author: federico_valeri
+---
+
+The Topic Operator enhances Strimzi's capabilities by allowing Kafka administrators to manage Kafka topics with Kubernetes custom resources.
+You can declare the desired topic state and the Topic Operator will take care of the reconciliation with the actual Kafka cluster state.
+
+The first generation of the Topic Operator is based on the Kafka Streams library, and uses ZooKeeper to gather topic metadata.
+We call it **Bidirectional Topic Operator (BTO)**, as it allows to reconcile changes coming from both Kubernetes and Kafka using a three-way merge logic.
```
scholzj, line 12: I think it would be worth to elaborate on this by adding simple examples -> 
* you create or change a topic in Kube -> it is applied to Kafka
* you create to change a topic in Kafka -> it is applied to Kube

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+    <figcaption><small>Fig 1. Unidirectional Topic Operator interactions.</small></figcaption>
+</figure>
+
+Users can create and dynamically change Kafka topic configurations by interacting with the `KafkaTopic` custom resources in real-time.
+A reconciliation loop compares the desired state in `KafkaTopic` resources with the current state of Kafka topics in Kafka.
+Actions are taken to align the actual topic configurations with the custom resource specifications.
+If the Kafka-side configuration is changed out-of-band (e.g. by using the Admin client, or Kafka shell scripts), those changes will eventually be reverted.
+
+The operator requires that a single `KafkaTopic` resource is used to manage a single topic in a single Kafka cluster.
+It detects cases where multiple resources are attempting to manage a Kafka topic using the same `spec.topicName`.
```
scholzj, line 45: This is the usual notation we use. We should stick with it.

```suggestion
It detects cases where multiple resources are attempting to manage a Kafka topic using the same `.spec.topicName`.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+Only the oldest `KafkaTopic` resource will be reconciled, while the other will fail with a `ResourceConflict` error.
+
+In addition to be able to pause reconciliations, the operator also provides a way to unmanage a Kafka topic using annotations.
+If the `KafkaTopic` is paused and we delete it, then it will be also deleted in Kafka.
+If the `KafkaTopic` is unmanaged and we delete it, then only the resource will be garbage collected by Kubernetes, but the topic will still exists in Kafka.
+
+As with previous implementation, we still do not support decreasing `spec.partitions` and changing `spec.replicas`.
```
scholzj, line 52:
```suggestion
As with previous implementation, we still do not support decreasing `.spec.partitions` and changing `.spec.replicas`.
```

Also, Kafka does not support a decreasing number of partitions. I think we should make that more clear by rephrasing this.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+While the former is not supported by Kafka itself, support for the latter could be added in the future.
+
+### Scalability
+
+The BTO is limited in term of scalability as it t only operates on one topic at a time, and it has to update a persistent store containing topic metadata.
```
scholzj, line 57:
```suggestion
The BTO is limited in terms of scalability as it only operates on one topic at a time, and it has to update a persistent store containing topic metadata.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+Instead, the UTO does not store topic metadata and it aims to be scalable in terms of the number of topics that it can operate on.
+
+When running Kafka operations, the UTO makes use of the request batching supported by the Kafka Admin client to get higher throughput for metadata operations.
+All `KafkaTopic` events are queued when received, and then processed in batches by a number of controller threads (currently, only a single thread is supported).
+If more than one event related to a single topic resource are found in the batch building, they are put back in the queue to be processed with the next batch.
+
+The following line graph confirms the BTO scalability issue, while the UTO can scale almost linearly.
+Using the UTO with batch size 500 and linger 1000, we were able to achieve a max reconciliation time of 5.3 seconds with 10k concurrent topic events.
```
scholzj, line 65: This does not seem to be in the chart. It is also not clear what `batch size 500 and linger 1000`. Maybe you can move this below the chart and incorporate it into the paragraph about these options?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+
+<figure>
+    <img src="/assets/images/posts/2023-10-23-uto-max-recon-time.png" height=350>
+    <figcaption><small>
+        Fig 2. Line graph comparing BTO and UTO max reconciliation times.<br/>
+        We used a 3-nodes cluster with default configurations, running on a local Minikube instance with 10 cores and 28 GB of memory.<br/>
+        The test driver application was running on the same machine, and the time spent on the queue was not included.
```
scholzj, line 72: What is the test driver here? Also, what does the queue time mean? I would expect the chart to show creation -> readiness or update -> readiness times which is what the user likely cares about.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+    </small></figcaption>
+</figure>
+
+You can tune the batching mechanism by setting `STRIMZI_MAX_QUEUE_SIZE` (default: 1024), `STRIMZI_MAX_QUEUE_SIZE` (default: 100), and `MAX_BATCH_LINGER_MS` (default: 100).
+If you exceed the configured max queue size, the UTO will print an error and then shutdown (Kubernetes will take care of restarting the pod).
+In that case, you can simply raise the max queues size to avoid the periodic operator restarts.
+
+### Race condition
+
+There is a known race condition happening when the operator and external applications try to change a topic configuration concurrently.
```
scholzj, line 82: You should clarify that you talk about changing it in Kafka here?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+That reconfiguration will either succeed, or fail.
+
+The recommendation here is to configure the Kafka cluster with `auto.create.topics.enable: false` in order to reduce the likelihood of hitting this problem.
+Unfortunately, disabling topic creation doesn't help with applications that use the Kafka Admin client to directly create required topics (i.e. Kafka Streams based applications).
+In this case, the application should be written to wait for topic existence, or use an init container to do the same.
+
+### Finalizers
+
+[Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers) are used by default to avoid missing topic deletion events when the UTO is not running.
+A common pitfall is that the namespace becomes stuck in a "terminating" state when you try to delete it without first removing all finalizers.
```
scholzj, line 97: I wonder if removing the finalizers is the right way to suggest. Should we instead suggest deleting the topics while the UTO is running? What you are suggesting here is great when you are deleting the namespace including the whole Kafka cluster and UTO. But not when you want to delete a topic while UTO has some outage.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+Given the way Kafka is implemented, there is no solution to that problem, but the UTO significantly reduces the likelihood of reconciliation failures.
+
+When the external application wins, the topics will be created using the default Kafka cluster configuration.
+In this case, the BTO simply fails the reconciliation, and you need to recover manually.
+Instead, the UTO always reconciles, which results in reconfiguration only when the default configuration differs from the spec.
+That reconfiguration will either succeed, or fail.
+
+The recommendation here is to configure the Kafka cluster with `auto.create.topics.enable: false` in order to reduce the likelihood of hitting this problem.
+Unfortunately, disabling topic creation doesn't help with applications that use the Kafka Admin client to directly create required topics (i.e. Kafka Streams based applications).
+In this case, the application should be written to wait for topic existence, or use an init container to do the same.
```
scholzj, line 92: This is not my experience. Do you have any numbers to back this up?

From my experience, the BTO often creates/updates the `KafkaTopic` with the values from Kafka and is happy with it. Only the values are usually different from what the user requested.

I would argue the TO does not improve this much and the conflict will be still there. If the topic is auto-created with default settings, the KafkaTopic might change it and lead to keys being mixed up in partitions etc.

So I wonder if we should instead aim at disabling auto-creation in the first place without suggesting UTO helps here.

----

Also, is Kafka Streams really an issue here? It creates its own topic and users typically don't create them with KafkaTopic resources. So I don't think I ever herd about conflict for a Streams topic.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+          value: "false"
+```
+
+### Upgrading from BTO to UTO
+
+The UTO is available as alpha feature since Strimzi 0.36.0, so we need to enable a feature gate in order to use it.
+[Strimzi 0.38.0](https://github.com/strimzi/strimzi-kafka-operator/releases/tag/0.38.0) adds Prometheus metrics and pause reconciliation feature.
+
+**This is not production ready yet, but the plan is to move it to beta in Strimzi 0.39.0, where it will be enabled by default.**
+**The UTO might require different CPU and memory resources, so we recommend to monitor and adjust accordingly.**
```
scholzj, line 133: I would either add specific (e.g. double the memory) or leave it out to be honest.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+            value: "false"
+
+# from standalone Deployment
+spec:
+  template:
+    spec:
+      containers:
+        - name: STRIMZI_USE_FINALIZERS
+          value: "false"
+```
```
scholzj, line 125: Adding the head of the resource (The `KAfka` CR with the basic metadata a the same for Deployment) will make this easier to understand to users what are you talking about.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+subscription.operators.coreos.com/<subscription-name> patched
+```
+
+At this point the Cluster Operator will restart, and then it will deploy the new Entity Operator containing the UTO.
+You can check the logs to confirm that the UTO is running fine.
+
+```sh
+$ kubectl logs $(kubectl get po -l strimzi.io/name=my-cluster-entity-operator -o name) -c topic-operator | "Returning from main"
+2023-10-23 08:32:25,60913 INFO  [main] TopicOperatorMain:148 - Returning from main()
+```
```
scholzj, line 158: Aren't there settings printed at startup that would be more useful to grep for?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+metadata:
+  name: uto-test
+  labels:
+    strimzi.io/cluster: my-cluster
+spec:
+  partitions: 1
+  replicas: 1
+  config:
+    min.insync.replicas: 1" | kubectl create -f -
+kafkatopic.kafka.strimzi.io/uto-test created
```
scholzj, line 174: Would the - at least for me - the more traditional form be more readable?

```suggestion
cat <<EOF | kubectl apply -f -
apiVersion: kafka.strimzi.io/v1beta2
kind: KafkaTopic
metadata:
  name: uto-test
  labels:
    strimzi.io/cluster: my-cluster
spec:
  partitions: 1
  replicas: 1
  config:
    min.insync.replicas: 1
EOF
```

It makes the YAML much more readable and easier to see and copy because it is well-aligned on its own lines.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+
+$ kubectl get kt uto-test -o yaml | yq '.status'
```
scholzj, line 176: Would `kubectl get kt uto-test -o jsonpath='{.status}'` be better? It does not need `yq` that is complicated because there are two of them and not all versions work the same. (hat said, it does not return YAML, so just a suggeston 🤷 )

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Downgrading from UTO to BTO
+
+Downgrading to BTO is as easy as disabling the feature gate.
+
+**Note that it may take some time to rebuild the internal state store if your Kafka cluster hosts many topics.**
+**Make sure to have adequate CPU and memory resources for that.**
+
+```sh
+$ kubectl set env deploy strimzi-cluster-operator STRIMZI_FEATURE_GATES=""
```
scholzj, line 228: If you set it to `-UnidirectionTopicOperator`, it would work even with 0.39 when it is enabled by default.

scholzj (commented):
I think the blog post is good. I left bunch of nits - but many of them are more for consideration rather than something you have to accept 😉.

The detailed chapters are good. But one thing I would probably improve is the introduction. I think you overestimate what the users know abotu BTO and UTO. I would spend more time to invest in explaining how BTO works and what is the difference in UTO. You know it, but many readers won't.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Yeah, the problem with this is that you can't align with other bash code (i.e. inside a for loop for example). With time, I get used to it and now I always use this form.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Unfortunately no. This is something we can improve.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Right, there is no difference between UTO and BTO with this issue.

A Streams application uses the admin client to create all required topics directly (CreateTopics API), and does not rely on auto topic creation.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Ok. Let me qualify this queue and add readiness timing on the tuning test without graph.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: I think it is clear if you continue to read few more lines where I talk about the UTO. I really don't want to give too much details about the 3way merge, as the main focus here is the new implementation.

fvaleri:
@scholzj it should now be ready for a 2nd pass. Thanks for all the feedback so far.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,312 @@
...
+Only the oldest resource will be reconciled, while the other resource will fail with a `ResourceConflict` error.
+
+We still do not support decreasing `.spec.partitions` and changing `.spec.replicas`.
+While the former is not supported by Kafka itself, support for the latter could be added in the future.
+
+In addition to be able to pause reconciliations, the operator now also provides a way to unmanage a Kafka topic using annotations.
+If the `KafkaTopic` is paused and we delete it, then it will be also deleted in Kafka.
+If the `KafkaTopic` is unmanaged and we delete it, then only the resource will be garbage collected by Kubernetes, but the topic will still exists in Kafka.
+
+In terms of requirements, the UTO assumes the following grants:
```
scholzj, line 91: Access rights would be more clear than _grants_.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,312 @@
...
+The BTO is limited in terms of scalability, as it only operates on one topic at a time, and it has to update a persistent store containing topic metadata.
+Instead, the UTO does not store topic metadata and it aims to be scalable in terms of the number of topics that it can operate on.
+The following line graph confirms the BTO scalability issue, while the UTO scales almost linearly.
+
+<figure>
+    <img src="/assets/images/posts/2023-10-23-uto-max-recon-time.png" height=350>
+    <figcaption><small>
+        Fig 2. Line graph comparing BTO and UTO max reconciliation times.<br/>
+        We used a 3-nodes cluster with default configurations, running on a local Minikube instance with 10 cores and 28 GB of memory.<br/>
+        The test driver application was running on the same machine, and the time spent on the event queue was not included.
```
scholzj, line 112: This is still misleading. When you move from event-driven architecture to batch architecture the per-event efficiency almost always improves. But what matters here is the _time-to-reaction_. How long does it take from the event happening till the reconciliation is complete. So this needs to include the time it was waiting in the queue or be removed. I would expect it to come out as a better number even with the queue - at least when running it at scale.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,312 @@
...
+All `KafkaTopic` events are queued when received, and then processed in batches by a number of controller threads (currently, only a single thread is supported).
+If more than one event related to a single topic resource are found in the batch building, they are put back in the queue to be processed with the next batch.
+
+You can tune the batching mechanism by setting `STRIMZI_MAX_QUEUE_SIZE` (default: 1024), `STRIMZI_MAX_BATCH_SIZE` (default: 100), and `MAX_BATCH_LINGER_MS` (default: 100).
+If you exceed the configured max queue size, the UTO will print an error and then shutdown (Kubernetes will take care of restarting the pod).
+In that case, you can simply raise the max queues size to avoid the periodic operator restarts.
+
+In another UTO test, we set max queue size to `MAX_INT`, max batch size to 200, batch linger ms to 500 and reconciliation interval to 10 seconds.
+With these settings, we were able to achieve a max reconciliation time of 2.5 seconds with 10k concurrent topic events.
+This translates to 3325 new `KafkaTopic` custom resources to become ready in about 2 minutes and 40 seconds.
```
scholzj, line 126: What exactly does this mean? Creating 10k Kubernetes resources normally takes some time. So are they pre-created when you start the operator? This would deserve explanation because users might try it ;-).

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,312 @@
...
+Unfortunately, disabling auto topic creation doesn't help with applications that use Kafka Admin client to directly create required topics (i.e. Kafka Streams apps).
+In this case, the application should be written to wait for topic existence, or use an init container to do the same.
+
+#### Finalizers
+
+[Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers) are used by default to avoid missing topic deletion events when the UTO is not running.
+If this happens, you end up with the `KafkaTopic` resource deleted in Kubernetes by garbage collection, and a Kafka topic still present in Kafka.
+
+It is recommended to delete `KafkaTopic` resources when the operator is running in order to trigger the required cleanups.
+A common pitfall is that the namespace becomes stuck in a "terminating" state when you try to delete it without first deleting all topics.
```
scholzj, line 146: Maybe like this?

```suggestion
A common pitfall is that an attempt to delete a namespace or Custom Resource Definition becomes stuck in a "terminating" state when you try to delete everything at once and the operator does not have time to delete all the `KafkaTopic` resources.
```

scholzj (commented):
Some more comments. As I will be traveling, feel free to merge it without my approval if everyone else approves it.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: I don't think that deleting CRDs is that common.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: I'm comparing apples to apples here. We only measure reconciliation time, which does not include queuing time. 

The aim here is to show that the reconciliation is faster and more predictable, which is something the user can verify using provided metrics. 

In the other test mentioned below, I'm also reporting the readiness time which includes queuing time.

That said I'm open to proposals here.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Will try to improve the description, but topics are not pre-created. This [test driver](https://github.com/fvaleri/examples/blob/main/misc/strimzi-load/src/main/java/it/fvaleri/example/Main.java#L77) first creates 5k topics and then you have 2.5k random update attempts and 2.5k random delete attempts. This test ends up with around 3.3k topics and I'm reporting the readiness time (2.4 min).

scholzj: Well, that does not sound like 10k concurrent events. If you create 5k new topics, how come only 3325 are created in the 2:40 interval? How quickly are full 5.000 topics created? Is the 2:40 interval really caused by the UTO or by the topic creation?

----

I get that a test like this can be helpful _internally_ to ensure the UTO has the expected performance after some changes. But I don't think this can be presented as a scalability numbers.

_posts/2023-10-23-unidirectional-topic-operator.md
scholzj: You might be comparing the same fruits. But it does not seem like apples but more like some fruit that nobody eats. What value does this number give me as a user? I don't care how long the reconciliation takes. I care about the time from the event happening in Kubernetes to the end of the reconciliation. As a user, I don't care that your reconciliation time is constant if the time in the queue is increasing to achieve that. This is important because you trade between more efficient reconciliation against time spent in queue. (The complete time is also what matters in the follow-up chapter where you talk about the race condition -> the window for it is defined by the time spent in queue and reconciliation.)

_posts/2023-10-23-unidirectional-topic-operator.md
scholzj: I think users who:
* Install the operator with `kubectl apply -f ...`
* Do something
* Try to uninstall the operator with `kubectl delete -f ...` on the same set of installation files and get stuck will be at least as common as users deleting the namespace.

fvaleri: That's true. Thanks for raising.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+When the external application wins, the topic will be created using the default Kafka cluster configuration.
+In this case, the UTO tries to reconcile to the desired state, and only fails when an incompatible change is detected (i.e. partition decrease).
+
+The recommendation here is to configure the Kafka cluster with `auto.create.topics.enable: false` in order to reduce the likelihood of hitting this problem.
+Unfortunately, disabling auto topic creation doesn't help with applications that use Kafka Admin client to directly create required topics (i.e. Kafka Streams apps).
+In this case, the application should be written to wait for topic existence, or use an init container to do the same.
+
+#### Finalizers
+
+[Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers) are used by default to avoid missing topic deletion events when the UTO is not running.
```
see-quick, line 143:
```suggestion
[Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers) are used by default to ensure that topic deletion events are not missed when the UTO is not running.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+If this happens, you end up with the `KafkaTopic` resource deleted in Kubernetes by garbage collection, and a Kafka topic still present in Kafka.
+
+It is recommended to delete `KafkaTopic` resources when the operator is running in order to trigger the required cleanups.
+A common pitfall is that an attempt to delete a namespace or Custom Resource Definition becomes stuck in a "terminating" state.
+This happens when you try to delete everything at once and the operator does not have time to delete all the `KafkaTopic` resources.
+In that case, you can simply remove finalizers from all `KafkaTopic` resources at once with the following command.
```
see-quick, line 149:
```suggestion
In such a scenario, you can remove the finalizers from all `KafkaTopic` resources simultaneously using the following command.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+$ kubectl -n <namespace> patch sub <subscription-name> --type merge -p '
+  spec:
+    config:
+      env:
+        - name: STRIMZI_FEATURE_GATES
+          value: "+UnidirectionalTopicOperator"'
+subscription.operators.coreos.com/<subscription-name> patched
+```
+
+At this point the Cluster Operator will restart, and then it will deploy the new Entity Operator containing the UTO.
```
see-quick, line 214:
```suggestion
At this point the Cluster Operator will restart, and then it will deploy the Entity Operator with the UTO included.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
+---
+layout: post
+title: "Unidirectional Topic Operator"
+date: 2023-10-23
+author: federico_valeri
+---
+
+The Topic Operator enhances Strimzi's capabilities by allowing Kafka administrators to manage Apache Kafka topics with Kubernetes custom resources (CRs).
```
see-quick, line 8:
```suggestion
The Topic Operator enhances Strimzi's capabilities by allowing Kafka administrators to manage Apache Kafka topics using Kubernetes custom resources (CRs).
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+You can declare the desired topic state and the Topic Operator will take care of the reconciliation with the actual Kafka cluster state.
```
see-quick, line 9:
```suggestion
You can declare the desired topic state, and the Topic Operator will reconcile it with the actual Kafka cluster state.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+In that case, you can simply raise the max queues size to avoid the periodic operator restarts.
+
+In another UTO test, we set max queue size to `MAX_INT`, max batch size to 200, batch linger ms to 500 and reconciliation interval to 10 seconds.
+This test driver creates 5k topics and then you have 2.5k random update attempts and 2.5k random delete attempts.
+With these settings, we were able to achieve a max reconciliation time of 2.5 seconds with 10k concurrent topic events.
+This means 3300 new `KafkaTopic` custom resources that are ready in about 2 minutes and 40 seconds.
+
+#### Race condition
+
+There is a known race condition happening when the operator and an external application try to change a topic configuration concurrently in Kafka.
```
see-quick, line 131:
```suggestion
There exists a known race condition when the operator and an external application attempt to change a topic configuration concurrently in Kafka.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+For example, it could be a GitOps pipeline creating the `KafkaTopic` resource while the application is doing the same during startup.
```
see-quick, line 132:
```suggestion
For instance, a GitOps pipeline might create the `KafkaTopic` resource while the application does the same during its startup.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+
+When the external application wins, the topic will be created using the default Kafka cluster configuration.
```
see-quick, line 134:
```suggestion
If the external application prevails, the topic is created using the default Kafka cluster configuration.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,314 @@
...
+        We used a 3-nodes cluster with default configurations, running on a local Minikube instance with 10 cores and 28 GB of memory.<br/>
+        The test driver application was running on the same machine, and the time spent on the event queue was not included.
+    </small></figcaption>
+</figure>
+
+When running Kafka operations, the UTO makes use of the request batching supported by the Kafka Admin client to get higher throughput for metadata operations.
+All `KafkaTopic` events are queued when received, and then processed in batches by a number of controller threads (currently, only a single thread is supported).
+If more than one event related to a single topic resource are found in the batch building, they are put back in the queue to be processed with the next batch.
+
+You can tune the batching mechanism by setting `STRIMZI_MAX_QUEUE_SIZE` (default: 1024), `STRIMZI_MAX_BATCH_SIZE` (default: 100), and `MAX_BATCH_LINGER_MS` (default: 100).
```
see-quick, line 120: Just an idea, which comes to my mind. I think it would be great to try to do some experiments with these parameters + (deploy 3 different clusters) and see what the results are and talk about it in the next blog post. WDYT?

see-quick (approved):
LGTM, just a few minor nits and one idea to consider.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Yes. I already did some experiments using a custom Java load test driver, but don't want to add too much to this presentation. A follow up blog post about tuning would be great, also considering that we are adding some [additional metrics](https://github.com/strimzi/strimzi-kafka-operator/pull/9216) that can be used for that.

tombentley (commented):
Thanks @fvaleri!

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Fair enough, I'm doing some more load tests and then will update that section.

fvaleri:
I replaced internal with end-to-end reconciliation time. Every test operation waits for readiness. Let me know what you think, and if you want more details.

scholzj (approved):

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,311 @@
+---
+layout: post
+title: "Unidirectional Topic Operator"
```
ppatierno, line 3: Maybe we can think of a better title? Not just the name of the operator you are going to talk about. Or even something like "Introduction to the new Unidirectional Topic Operator" or something better.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,311 @@
...
+date: 2023-10-23
+author: federico_valeri
+---
+
+The Topic Operator enhances Strimzi's capabilities by allowing Kafka administrators to manage Apache Kafka topics using Kubernetes custom resources (CRs).
+You can declare the desired topic state, and the Topic Operator will reconcile it with the actual Kafka cluster state.
```
ppatierno, line 9: Maybe "desire topic configuration" is better? Not sure talking about "topic state" makes sense, or what did you have in mind?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,311 @@
...
+```
+
+The requirement here is that a single `KafkaTopic` custom resource is used to manage a single topic in a single Kafka cluster.
+From the example above, you can see that the resource has a label `strimzi.io/cluster` with the name of the target cluster.
+The operator communicates with the Kafka cluster and ensures that the specified Kafka topic is created or updated according to the desired state.
+It can be deployed standalone, or along with your Kafka cluster using the `Kafka` custom resource.
+
+The first generation of the Topic Operator is based on the Kafka Streams library, and uses ZooKeeper to gather topic metadata.
+We call it **Bidirectional Topic Operator (BTO)**, as it allows to reconcile changes coming from both Kubernetes and Kafka using a three-way merge logic.
+In order to do this, it has to keep an internal topic store, which is the source-of-truth used to determine the necessary changes for synchronization.
```
ppatierno, line 73: The last two sentences are a repetion of what you already said before. They are almost exactly the same sentences.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,311 @@
...
+- list partition reassignments
+- describe broker configs
+
+#### Scalability
+
+The BTO is limited in terms of scalability, as it only operates on one topic at a time, and it has to update a persistent store containing topic metadata.
+Instead, the UTO does not store topic metadata and it aims to be scalable in terms of the number of topics that it can operate on.
+
+When running Kafka operations, the UTO makes use of the request batching supported by the Kafka Admin client to get higher throughput for metadata operations.
+All `KafkaTopic` events are queued when received, and then processed in batches by a number of controller threads (currently, only a single thread is supported).
```
ppatierno, line 107: What's a `KafkaTopic` event? Could you clarify please? I could imagine it's about the creation/update/deletion of the custom resource, or? Can you clarify it?

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Duplication removed.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: There is no old Unidirectional TO. What about "Introducing the Unidirectional Topic Operator"?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+- list partition reassignments
+- describe broker configs
+
+#### Scalability
+
+The BTO is limited in terms of scalability, as it only operates on one topic at a time, and it has to update a persistent store containing topic metadata.
+Instead, the UTO does not store topic metadata and it aims to be scalable in terms of the number of topics that it can operate on.
+
+When running Kafka operations, the UTO makes use of the request batching supported by the Kafka Admin client to get higher throughput for metadata operations.
+All `KafkaTopic` events (creations, updates, deletes) are queued when received, and then processed in batches by a number of controller threads (currently, only a single thread is supported).
```
ppatierno, line 106: What do you think about this?

```suggestion
All `KafkaTopic` related events (custom resource creations, updates, deletes) are queued when received, and then processed in batches by a number of controller threads (currently, only a single thread is supported).
```

ppatierno (approved):
LGTM overall. I left just one suggestion.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Sounds good, thanks.

fvaleri:
I'm planning to run the load tests again after some code changes, so please give me some more time.

ppatierno:
> I'm planning to run the load tests again after some code changes, so please give me some more time.

Sounds goot to me.
It would be great waiting for a review by @PaulRMellor as well.

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+The requirement here is that a single `KafkaTopic` custom resource is used to manage a single topic in a single Kafka cluster.
+From the example above, you can see that the resource has a label `strimzi.io/cluster` with the name of the target cluster.
+The operator communicates with the Kafka cluster and ensures that the specified Kafka topic is created or updated according to the desired state.
+It can be deployed standalone, or along with your Kafka cluster using the `Kafka` custom resource.
+
+The first generation of the Topic Operator is based on the Kafka Streams library, and uses ZooKeeper to gather topic metadata.
+We call it **Bidirectional Topic Operator (BTO)**, as it allows to reconcile changes coming from both Kubernetes and Kafka using a three-way merge logic.
+In order to do this, it has to keep an internal topic store, which is the source-of-truth used to determine the necessary changes for synchronization.
+If the Kafka-side configuration is changed out-of-band (e.g. by using the Admin client, or Kafka shell scripts), those changes will eventually be reverted.
```
PaulRMellor, line 73: I thought this was okay with BTO and reversion was with UTO?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+The new generation of the Topic Operator is called **Unidirectional Topic Operator (UTO)**.
+The UTO improves on scalability and maintenance, but also provides support for Kafka clusters running in KRaft mode.
+It is fully backwards compatible with the BTO, but it will only reconcile topic events in one direction, from Kubernetes to Kafka.
+Your `KafkaTopic` custom resources will be the source-of-truth, like with all the other operators.
```
PaulRMellor, line 78:
```suggestion
The **Unidirectional Topic Operator (UTO)** is the latest generation of Topic Operator, which improves on the BTO in terms of topic maintenance and scalability, but also provides support for Kafka clusters running in KRaft mode.
As the UTO only reconciles topic events from Kubernetes to Kafka,`KafkaTopic` custom resources become the single source-of-truth.
```

PaulRMellor: We introduced the UTO once. Maybe we can reduce as suggested here^  I thought we could introduce maintenance and scalability here and then have separate subsections...

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+The operator detects cases where multiple `KafkaTopic` resources are attempting to manage a Kafka topic using the same `.spec.topicName`.
```
PaulRMellor, line 80:
```suggestion

#### Topic maintenance

The new UTO can detect cases where multiple `KafkaTopic` resources are attempting to manage a Kafka topic using the same `.spec.topicName`.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+Only the oldest resource will be reconciled, while the other resource will fail with a `ResourceConflict` error.
```
PaulRMellor, line 81:
```suggestion
Only the oldest resource is reconciled, while the other resource fail with a `ResourceConflict` error.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+We still do not support decreasing `.spec.partitions` and changing `.spec.replicas`.
```
PaulRMellor, line 83:
```suggestion
The UTO still does not support decreasing `.spec.partitions` and changing `.spec.replicas`.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+While the former is not supported by Kafka itself, support for the latter could be added in the future.
+
+In addition to be able to pause reconciliations, the operator now also provides a way to unmanage a Kafka topic using annotations.
```
PaulRMellor, line 86:
```suggestion
In addition to being able to pause reconciliations, the UTO also provides a way to unmanage a Kafka topic using annotations.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+- create topics
+- create partitions
+- delete topics
+- list partition reassignments
+- describe broker configs
+
+#### Scalability
+
+The BTO is limited in terms of scalability, as it only operates on one topic at a time, and it has to update a persistent store containing topic metadata.
+Instead, the UTO does not store topic metadata and it aims to be scalable in terms of the number of topics that it can operate on.
```
PaulRMellor, line 103:
```suggestion
By contrast, the UTO does not store topic metadata and it aims to be scalable in terms of the number of topics that it can operate on.
```

PaulRMellor: Not quite sure what _aims to be_ means here? Do we need that?  Should we say anything about limits to the number of topics it can operate on if there are any?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+When running Kafka operations, the UTO makes use of the request batching supported by the Kafka Admin client to get higher throughput for metadata operations.
+All `KafkaTopic` related events (custom resource creations, updates, deletes) are queued when received, and then processed in batches by a number of controller threads (currently, only a single thread is supported).
+If more than one event related to a single topic resource are found in the batch building, they are put back in the queue to be processed with the next batch.
```
PaulRMellor, line 107:
```suggestion
If more than one event related to a single topic resource is found in the batch building, the events are put back in the queue to be processed with the next batch.
```

PaulRMellor: Not sure if that's what we want to say here. Just fixing disagreement with _more than one event_

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+You can tune the batching mechanism by setting `STRIMZI_MAX_QUEUE_SIZE` (default: 1024), `STRIMZI_MAX_BATCH_SIZE` (default: 100), and `MAX_BATCH_LINGER_MS` (default: 100).
```
PaulRMellor, line 109: We haven't documented these. Not sure we should?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+<figure>
+    <img src="/assets/images/posts/2023-10-30-uto-recon-graph.png" height=350>
+    <figcaption><small>
+        Fig 2. Line graph comparing BTO and UTO end-to-end reconciliation time.<br/>
+        Environment: Strimzi 0.38.0, 3-nodes cluster running on Minikube.
+    </small></figcaption>
+</figure>
+
+The above line graph shows that the UTO is faster and scale much better with a mixed workload of topic operations, which is similar to what happens in busy clusters.
```
PaulRMellor, line 121:
```suggestion
The above line graph shows that the UTO is faster and scales much better than the BTO with a mixed workload of topic operations, which is similar to what happens in busy clusters.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+The only configuration changes we did were: `.spec.entityOperator.topicOperator.reconciliationIntervalSeconds: 10` (both), and `MAX_BATCH_LINGER_MS=10` (UTO only).
```
PaulRMellor, line 122:
```suggestion
The only configuration changes we made were: `.spec.entityOperator.topicOperator.reconciliationIntervalSeconds: 10` (both), and `MAX_BATCH_LINGER_MS=10` (UTO only).
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+#### Race condition
+
+There exists a race condition when the operator and an external application attempt to change a topic configuration concurrently in Kafka.
+For instance, a GitOps pipeline might create a topic using a `KafkaTopic` resource, while the application tries to create the same topic during its startup.
+
+If the external application prevails, the topic is created using the default Kafka cluster configuration.
+In this case, the UTO tries to reconcile to the desired state, and only fails when an incompatible change is detected (i.e. partition decrease).
+
+The recommendation here is to configure the Kafka cluster with `auto.create.topics.enable: false` in order to reduce the likelihood of hitting this problem.
```
PaulRMellor, line 133:
```suggestion
The recommendation when using the UTO is to configure the Kafka cluster with `auto.create.topics.enable: false` in order to reduce the likelihood of hitting this problem.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+Unfortunately, disabling auto topic creation doesn't help with applications that use Kafka Admin client to directly create required topics (i.e. Kafka Streams apps).
```
PaulRMellor, line 134:
```suggestion
Unfortunately, disabling auto topic creation doesn't help with applications that use the Kafka Admin client to create topics directly, such as with Kafka Streams apps.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+In this case, the application should be written to wait for topic existence, or use an init container to do the same.
```
PaulRMellor, line 135:
```suggestion
In this case, the application should be written to wait for a topic to exist or use an init container to do the same.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+#### Finalizers
+
+[Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers) are used by default to ensure that topic deletion events are not missed when the UTO is not running.
+If this happens, you end up with the `KafkaTopic` resource deleted in Kubernetes by garbage collection, and a Kafka topic still present in Kafka.
+
+It is recommended to delete `KafkaTopic` resources when the operator is running in order to trigger the required cleanups.
+A common pitfall is that an attempt to delete a namespace or Custom Resource Definition becomes stuck in a "terminating" state.
+This happens when you try to delete everything at once and the operator does not have time to delete all the `KafkaTopic` resources.
+In such a scenario, you can remove the finalizers from all `KafkaTopic` resources simultaneously using the following command.
```
PaulRMellor, line 145:
```suggestion
In such a scenario, you can remove the finalizers from all `KafkaTopic` resources simultaneously using the following command:
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+```sh
+$ kubectl get kt -o yaml | yq 'del(.items[].metadata.finalizers[])' | kubectl apply -f -
+kafkatopic.kafka.strimzi.io/topic-3 configured
+kafkatopic.kafka.strimzi.io/topic-1 configured
+kafkatopic.kafka.strimzi.io/topic-2 configured
+```
+
+In development and test environment you can disable the use of finalizers with the following configuration.
```
PaulRMellor, line 154:
```suggestion
In development and test environments, you can disable the use of finalizers with the following configuration:
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+  template:
+    spec:
+      containers:
+        - name: STRIMZI_USE_FINALIZERS
+          value: "false"
+```
+
+### Upgrading from BTO to UTO
+
+The UTO is available as alpha feature since Strimzi 0.36.0, so we need to enable a feature gate in order to use it.
```
PaulRMellor, line 187:
```suggestion
The UTO has been available as alpha feature since Strimzi 0.36.0, so you need to enable a feature gate in order to use it.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+[Strimzi 0.38.0](https://github.com/strimzi/strimzi-kafka-operator/releases/tag/0.38.0) adds Prometheus metrics and pause reconciliation feature.
```
PaulRMellor, line 188:
```suggestion
[Strimzi 0.38.0](https://github.com/strimzi/strimzi-kafka-operator/releases/tag/0.38.0) enhances the UTO with by adding Prometheus metrics and a pause reconciliation feature.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+**The plan is to move it to beta in Strimzi 0.39.0, where it will be enabled by default.**
+**Tests show that UTO consumes approximately 40% less CPU and 30% less memory than the BTO.**
```
PaulRMellor, line 191: This seems a bit out of context here. would it be better as a note with the scalability findings?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+```sh
+$ kubectl set env deploy strimzi-cluster-operator STRIMZI_FEATURE_GATES="+UnidirectionalTopicOperator"
+deployment.apps/strimzi-cluster-operator updated
```
PaulRMellor, line 195: maybe move this up to sit under _so we need to enable a feature gate in order to use it._?

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+At this point the Cluster Operator will restart, and then it will deploy the Entity Operator with the UTO included.
+You can check the logs to confirm that the UTO is running fine.
+
+```sh
+$ kubectl logs $(kubectl get po -l strimzi.io/name=my-cluster-entity-operator -o name) -c topic-operator | "Returning from main"
+2023-10-23 08:32:25,60913 INFO  [main] TopicOperatorMain:148 - Returning from main()
+```
+
+We can also try to create a new test topic and see if it is reconciled.
```
PaulRMellor, line 218:
```suggestion
Verify the UTO installation by creating a new test topic and seeing if it is reconciled:
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+$ kubectl exec my-cluster-kafka-0 -- bin/kafka-topics.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --describe --topic uto-test
+Topic: uto-test	TopicId: gWrR6uDrSNSSfu_ubGndCg	PartitionCount: 1	ReplicationFactor: 1	Configs: min.insync.replicas=1,message.format.version=3.0-IV1
+	Topic: uto-test	Partition: 0	Leader: 1	Replicas: 1	Isr: 1
+	
+$ kubectl delete kt uto-test
+kafkatopic.kafka.strimzi.io "uto-test" deleted
+```
+
+Optionally, you can delete the internal topics created by the BTO, which are not used anymore.
```
PaulRMellor, line 245:
```suggestion
Optionally, you can delete the internal topics created by the BTO, which are not used anymore:
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+```sh
+$ kubectl delete $(kubectl get kt -o name | grep strimzi-store-topic)
+kafkatopic.kafka.strimzi.io "strimzi-store-topic---effb8e3e057afce1ecf67c3f5d8e4e3ff177fc55" deleted
+
+$ kubectl delete $(kubectl get kt -o name | grep strimzi-topic-operator)
+kafkatopic.kafka.strimzi.io "strimzi-topic-operator-kstreams-topic-store-changelog---b75e702040b99be8a9263134de3507fc0cc4017b" deleted
+```
+
+Finally, we can also choose to unmanage the Kafka internal topics, without deleting them inside the Kafka cluster.
```
PaulRMellor, line 255:
```suggestion
Finally, you can also choose to unmanage the Kafka internal topics, without deleting them inside the Kafka cluster.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+Here I'm just showing the `__consumer_offsets`, but the procedure is the same for the other internal topics.
```
PaulRMellor, line 256:
```suggestion
Here I'm just showing the `__consumer_offsets` topic, but the procedure is the same for the other internal topics:
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+### Downgrading from UTO to BTO
+
+Downgrading to BTO is as easy as disabling the feature gate.
+
+**Note that it may take some time to rebuild the internal state store if your Kafka cluster hosts many topics.**
+**Make sure to have adequate CPU and memory resources for that.**
+
+```sh
+$ kubectl set env deploy strimzi-cluster-operator STRIMZI_FEATURE_GATES="-UnidirectionTopicOperator"
+deployment.apps/strimzi-cluster-operator env updated
```
PaulRMellor, line 282: Again, maybe we want to bring this up to sit under _Downgrading to BTO is as easy as disabling the feature gate._

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+```
+
+At this point the Cluster Operator will restart, and then will deploy the new Entity Operator containing the BTO.
```
PaulRMellor, line 285:
```suggestion
At this point the Cluster Operator restarts, and then deploy the new Entity Operator containing the BTO.
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+You can check the logs to confirm that the BTO is running fine.
```
PaulRMellor, line 286:
```suggestion
You can check the logs to confirm that the BTO is running fine:
```

_posts/2023-10-23-unidirectional-topic-operator.md
```diff
@@ -0,0 +1,310 @@
...
+
+```sh
+$ kubectl logs $(kubectl get po -l strimzi.io/name=my-cluster-entity-operator -o name) -c topic-operator | grep "Session deployed"
+2023-10-23 08:59:51,95138 INFO  [vert.x-eventloop-thread-1] Main:70 - Session deployed
+```
+
+After that, you should see all your topics, including the auto-created internal topics.
```
PaulRMellor, line 293:
```suggestion
After that, you should see all your topics, including the auto-created internal topics:
```

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: I think it's better to warn users before the command.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Wrong position, fixed. Thanks.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Not having to manage persistent state is important on the scalability side. It is hard to put some limit here. For example, Kafka doesn't provide one.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Not exactly. Let me improve that.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: I think we should, because they may be useful to fine tune the operator in some use cases.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: Well, I put that here because users may want to resize resources after the upgrade.

_posts/2023-10-23-unidirectional-topic-operator.md
fvaleri: I think it's better to warn users before the command.

fvaleri:
I think this is now ready for prime time.

PaulRMellor:
> I think this is now ready for prime time.

@fvaleri -- just one thing. should Downgrading from UTO to BTO say anything about removing/disabling finalizers if they are being used?

fvaleri:
> > I think this is now ready for prime time.
> 
> @fvaleri -- just one thing. should Downgrading from UTO to BTO say anything about removing/disabling finalizers if they are being used?

This is a very good point Paul. I added a note in that paragraph.

tombentley (approved):
Thanks @fvaleri !

PR #403, opened 2023-12-12
docs(web pages): updates to the ToC and general navigation improvements

PaulRMellor:
**Updates to doc web pages**

- Removes the resizeable ToC
- Adds a new ToC switch to turn the ToC on/off
- Reduces the size of the ToC text to put more titles on a single line
- Adds a dynamic "Top" button (mainly for mobile views to return to the ToC)
- Sets a more readable weight to the text


scholzj (approved):
Nice work. Thanks.

fvaleri (approved):
This is much better IMO, and there is no delay on window resize. Thanks!

ppatierno (approved):
Nice work Paul! Just saying that on my side, if you make the browser window smaller or you use a mobile phone, the TOC cannot be disabled but just stay at the top of the page as before. Is it normal?

PaulRMellor:
> Nice work Paul! Just saying that on my side, if you make the browser window smaller or you use a mobile phone, the TOC cannot be disabled but just stay at the top of the page as before. Is it normal?

Thanks Paolo. At smaller widths the side ToC disappears because of the width restrictions and you get the ToC (a different one) at the top. It's always been like that, and as it doesn't get in the way of the text there's no point in removing it like the side ToC. But the new "Top" button lets you go straight back to the top to see it again.

PR #407, opened 2024-01-10
docs(style): styles for large tables in the docs

PaulRMellor:
**Documentation content update**

Moves the style settings for large tables so that they are local and affect only documentation tables
The CSS properties  add a scroll bar to a table if the content overflows.
Previously, this was being set in the global CSS, which impacted the downloads page 

* [x] Typo/minor fix
* [ ] Other

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -51,7 +51,7 @@
   /* adds a horizontal scroll bar to any doc tables that expand beyond the container */
   table {    
     overflow-x: auto !important;
-    display:block;
+    display: block;
 }
 </style>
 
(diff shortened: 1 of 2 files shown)


scholzj (approved):

jankalinic (commented):
I think the display attribute will still cause trouble in the download section, if the display: block is inherited from this layout. If not then it's ok.

PaulRMellor:
> I think the display attribute will still cause trouble in the download section, if the display: block is inherited from this layout. If not then it's ok.

Thanks @jankalinic -- i checked it locally and the tables are fine. It doesn't inherit from this position.

PR #412, opened 2024-02-08
Add blog post about Kafka Tiered Storage

fvaleri:
This is a deep dive blog post about Kafka Tiered Storage. The objective is not have another tutorial, but to give some insights into the motivations, design, and implementation.

Draft waiting for some integrations by @showuon.


fvaleri:
Currently, we don't support this feature, so this may not be the best place to publish this.

PR #421, opened 2024-03-19
Fix style for bold/strong text

im-konge:
### Type of change

- Fix

### Description

This PR fixes issue with the style of `<b>`/`<strong>` elements, as the text wasn't highlighted at all.
The reason is that the style of those two tags were overwritten by style of other "parent" elements -> `<p>`, `<ul>`, .etc.

For not inheriting the `font-weight` for `b, strong` elements, I added explicit style for them, which fixes the issue.

Before:

image

After:

image

### DIFF
--- a/_sass/core/global.scss
+++ b/_sass/core/global.scss
@@ -117,6 +117,10 @@ p.byline {
   line-height: 1.8rem;
 }
 
+b,strong {
+  font-weight: bold;
+}
+
 a {
   line-height: 1.5rem;
   font-weight: 400;


_sass/core/global.scss
```diff
@@ -117,6 +117,10 @@ p.byline {
   line-height: 1.8rem;
 }
 
+b,strong {
+  font-weight: bold;
+}
+
```
scholzj, line 123: Why does it need to be added? It sounds like this is probably just hiding some other styling which you hard-override.

im-konge: other elements has other `font-weight` set there
for example `p`:
```
p {
  font-size: 1rem;
  line-height: 1.5rem;
  font-weight: 100;
}
```
so I guess that other sub-elements that are used as children are inheriting this `font-weight`

scholzj: Hmm 🤔

im-konge: And also for the whote `body, html` is set the `font-weight`:

```
html,
body,
*
{
  font-size: 100%;
  font-weight: 100;
  font-family: "Nunito", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif;
}
```

let me try to remove the `*`, but I'm a bit afraid that it will break something 😄

scholzj: I don't know ... better not poke that :-o

im-konge: Yeah even without the wildcard it doesn't work 🤷

scholzj (approved):
I guess it is ok if you check that it didn't broke anything anywhere.

im-konge:
It changed the blogposts where we are using `<b>` or Markdown's `** **`

But also `404 Not Found` page:
image

The YAML examples, code snippets, and everything else seems fine.

im-konge:
image
image

That's what I found that differs now

scholzj:
I guess those changes are fine 🤷

im-konge:
So I see also changes in the code snippets f.e. here:

Before:
image

After:
image

But from the documentation it's highlighted, so I guess that's fine as well

scholzj:
That seems wrong. Also, did it had different color before?

im-konge:
Yes, when you take a look into the documentation, there is this:

```
----
apiVersion: {KafkaApiVersion}
kind: Kafka
metadata:
  name: my-cluster
spec:
  #...
  entityOperator:
    topicOperator: {}
    *userOperator: {}*
----
```
which is then transformed to something like this

```
<pre class="highlightjs highlight"><code data-lang="yaml" class="language-yaml hljs">apiVersion: kafka.strimzi.io/v1beta1
kind: Kafka
metadata:
  name: my-cluster
spec:
  #...
  entityOperator:
    topicOperator: {}
    <strong>userOperator: {}</strong></code></pre>
</div>
```

the color is not dark grey (`#444`) as in other code blocks, but black (`#000`) based on:

```
* {
  color: $black;
  line-height: 1.5rem;
}
```

but that's there for a long time. My change just did what is in the documentation -> it made the text bold

im-konge:
If you check f.e. this blogpost: https://strimzi.io/blog/2024/01/03/prevent-broker-scale-down-if-containing-paritition-replicas/ 
it seems that each code snippet is styled and generated from the markdown differently.

So maybe that should be checked as well?

scholzj:
> Yes, when you take a look into the documentation, there is this:

Ah, ok. It makes of course sense if it is marked in the docs as bold.

im-konge:
> Ah, ok. It makes of course sense if it is marked in the docs as bold.

I can check the styles for that and maybe change to have it bold and gray instead of black or something.

see-quick (approved):
If this does not break anything. LGTM

Frawless (approved):

PaulRMellor (approved):
The bold looks better on the code now. There' a faint bold there presently. I'm sure the CSS could be streamlined to make these changes easier.

PR #472, opened 2025-02-07
docs(docs): adds links to content related to drain cleaner and access operator

PaulRMellor:
**Documentation page**

Adds links to the Documentation page for content related to the Drain Cleaner and Access Operator

* [ ] Typo/minor fix
* [ ] Other

### DIFF
--- a/_data/releases.yaml
+++ b/_data/releases.yaml
@@ -695,3 +695,7 @@ bridge:
   - version: 0.12.2
   - version: 0.12.1
   - version: 0.12.0
+drainCleaner:
+  - version: 1.2.0
+accessOperator:
+  - version: 0.1.0
--- a/_includes/documentation/documentation-version-band.html
+++ b/_includes/documentation/documentation-version-band.html
@@ -18,6 +18,18 @@
         <a href="{{site.baseurl}}/docs/bridge/latest/">Strimzi Kafka Bridge documentation</a> <a href="/docs/bridge/latest/full.html" target="_blank"><i class="fas fa-external-link-alt"></i></a>
       </p>
       <p>
+        <strong class="text-caps">Strimzi Drain Cleaner {{site.data.releases.drainCleaner[0].version}}</strong>
+        <br/>
+        <a href="{{site.baseurl}}/docs/operators/latest/deploying.html#assembly-drain-cleaner-str">Evicting pods with the Strimzi Drain Cleaner</a> <a href="/docs/operators/latest/full/deploying.html#assembly-drain-cleaner-str" target="_blank"><i class="fas fa-external-link-alt"></i></a>
+      </p>
+
+      <p>
+        <strong class="text-caps">Strimzi Access Operator {{site.data.releases.accessOperator[0].version}}</strong>
+        <br/>
+        <a href="{{site.baseurl}}/docs/operators/latest/deploying.html#assembly-access-operator-str">Managing client connections with the Access Operator</a> <a href="/docs/operators/latest/full/deploying.html#assembly-access-operator-str" target="_blank"><i class="fas fa-external-link-alt"></i></a>
+      </p>
+
+      <p>
         <strong class="text-caps">Quickstarts</strong>
         <br/>
         Try the <a href="{{site.baseurl}}/quickstarts/">quickstarts</a> for quick and easy installations on different development environments.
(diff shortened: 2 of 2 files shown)


_data/releases.yaml
```diff
@@ -695,3 +695,7 @@ bridge:
   - version: 0.12.2
   - version: 0.12.1
   - version: 0.12.0
+drainCleaner:
+  - version: 1.2.0
+accessOperator:
+  - version: 0.1.0
```
im-konge, line 701: I think we have `0.1.1` released

scholzj: We do have 0.1.1 out. But the Operator 0.45 docs and installation file have 0.1.0 in them as this is something we update only with next Operator release. So I guess this is a bit confusing 🤔. Not sure what is the best approach here. Should we have the link without mentioning any exact version and have the exact version in downloads?

im-konge: Yeah sorry, I didn't realize that we have the `0.1.0` in the installation files 🙂 
But as you mentioned, I would maybe keep it without the version. For Bridge, we have a separate section just for it, right? That's why it's okay to keep the version there, even if the version is not part of the Operators installation files.

katheris: I understand that Bridge can run against any Kafka, so it makes sense to me that it would have it's own link. I would propose a different approach. What if instead of adding links to Drain Cleaner and Access Operator here, which are only useful if you are using the Cluster Operator, we made them more prominent in the Cluster Operator docs pages. So e.g. mention them specifically in the overview and look at how they can be more easily found in the deploying and upgrading page?

ppatierno: Yes, the bridge can run against any Kafka and it's not relegated to Kubernetes but also bare metal and VMs so it deserves its own link imho. Regarding drain cleaner and access operator, finding a way to highlight them in the doc is not that simple, because they could be easily lost but maybe it's a good approach.

ppatierno: Where would you add this piece?

PaulRMellor: On the Documentation page. Above the guide links.

PaulRMellor:
In the last commit, I added an introduction to the Documentation page that introduces the operators and describes where to find information on them. I've made a couple more changes too: 

- Replaced the white text on blue background with dark text on white background for accessibility reasons. 
- Updated the titles for the deploying guide from version 0.28, as it's been Deploying and Managing since then. 
- Consolidated the content for the doc page into a single file
- Made the font colour for the doc headings consistent.

@katheris -- To support the new introduction, I'll look to make the available operators more obvious in the Overview too.

_includes/documentation/documentation-version-band.html
```diff
@@ -42,6 +61,14 @@ <h3>Strimzi {{site.data.releases.operator[0].version}} documentation (latest sta
...
       <a href="{{site.baseurl}}/docs/bridge/in-development/">Strimzi Kafka Bridge documentation</a> <a href="/docs/bridge/in-development/full/bridge.html" title="Full documentation (New Window)" target="_blank"><i class="fas fa-external-link-alt"></i></a>
       </p>
+
+      <p>
+        <strong class="text-caps">Strimzi Access Operator</strong>
+        <br/>
+        <a href="{{site.baseurl}}/docs/operators/in-development/deploying.html#assembly-access-operator-str">Managing client connections with the Access Operator</a> <a href="/docs/operators/in-development/full/deploying.html#assembly-access-operator-str" target="_blank"><i class="fas fa-external-link-alt"></i></a>
+        <a href="{{site.baseurl}}/docs/operators/in-development/deploying.html#assembly-drain-cleaner-str">Evicting pods with the Strimzi Drain Cleaner</a> <a href="/docs/operators/in-development/full/deploying.html#assembly-drain-cleaner-str" target="_blank"><i class="fas fa-external-link-alt"></i></a>
+        
+      </p>
```
katheris, line 71: How come there isn't an in-development link for the drain cleaner?

katheris (commented):
I like the new paragraph explaining about the different operators.

I think if we are going to add the drain cleaner and access operator as links here we should also include the topic and user operators.

Also the access operator docs seem to be under the "Managing TLS certificates" section, I think it should be a separate section

_includes/documentation/documentation-version-band.html
PaulRMellor: I didn't push up the cleaned up links. I'd removed them.

PaulRMellor:
> I like the new paragraph explaining about the different operators.
> 
> I think if we are going to add the drain cleaner and access operator as links here we should also include the topic and user operators.
> 
> Also the access operator docs seem to be under the "Managing TLS certificates" section, I think it should be a separate section

Thanks @katheris. The links weren't meant to be there. I didn't push up the cleaned-up version. So now we have the intro that describes the operators available and where to find information on them. If that's enough for this page, I'll update the Overview to have a specific section for each one.

ppatierno (approved):

scholzj (changes requested):
You should not rename the documentation to different name than it has in the release. It should use the same name as the documentation has. The `Deploying and Managing` name is used only from Strimzi 0.34.0 on.

PaulRMellor:
> You should not rename the documentation to different name than it has in the release. It should use the same name as the documentation has. The `Deploying and Managing` name is used only from Strimzi 0.34.0 on.

I can change that. We dropped the Using Guide at 0.27.1. I thought it made sense to switch the name at 0.28

scholzj:
> > You should not rename the documentation to different name than it has in the release. It should use the same name as the documentation has. The `Deploying and Managing` name is used only from Strimzi 0.34.0 on.
> 
> I can change that. We dropped the Using Guide at 0.27.1. I thought it made sense to switch the name at 0.28

I guess we called it `Deploying and Upgrading` in between. The docs are under that name in the release files as well. So I think we should keep the matching name to avoid any confusion.

PR #478, opened 2025-03-20
Phased Strimzi upgrade post

tomncooper:
### Type of change

* [ ] New blog post

This PR adds a blog post covering an example Strimzi deployment model which allows multiple Strimzi versions to be installed concurrently and uses the custom resource selector env var to govern which Strimzi instances manage which Kafka clusters. This allows infra teams to selectively upgrade Kafka clusters to new Strimzi versions in a controlled way with easy rollback.

This example uses [Kustomize](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/) to alter the base install resources to make sure they can live together on the same cluster. An [example repository](https://github.com/tomncooper/strimzi-phased-upgrade) is provided which shows the full deployment model and contains scripting and templates for adding new Strimzi versions.

This blog uses [mermaid](https://mermaid.js.org/) diagrams (written inline in the markdown) so adds the option to load the mermaid.js script by specifying `mermaid: true` in the markdown front-matter.

### DIFF
--- a/_posts/2025-04-10-phased-strimzi-upgrade-example.md
+++ b/_posts/2025-04-10-phased-strimzi-upgrade-example.md
@@ -1,7 +1,7 @@
 ---
 layout: post
 title: "Phased upgrades of Strimzi managed Kafka fleets"
-date: 2025-04-09
+date: 2025-04-10
 author: tom_cooper
 mermaid: true
 ---


_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
+---
+layout: post
+title: "Phased upgrades of Strimzi managed Kafka fleets"
+date: 2025-03-20
+author: tom_cooper
+---
+
+## TL;DR
```
scholzj, line 8: I would remove this title, it seems silly and unnecessary. All it does is it provides a simple introduction as all out other blog posts.

_includes/site-scripts.html
```diff
@@ -2,3 +2,13 @@
...
+<script src="https://cdn.jsdelivr.net/npm/mermaid@11.5.0/dist/mermaid.min.js"></script>
+<script>
+  $(document).ready(function () {
+    mermaid.initialize({
+      startOnLoad:true,
+      theme: "default",
+    });
+    window.mermaid.init(undefined, document.querySelectorAll('.language-mermaid'));
+  });
+</script>
```
scholzj, line 14: Is this something we really want in all pages? Isn't it better to just use a picture like everyone else?

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+When you upgrade an instance of the Strimzi Cluster Operator you will force **all** the Kafka clusters that it manages to roll. 
+For large fleets this could be quite disruptive.
+It would be nice to be able to decide which Kafka clusters upgrade to the new Strimzi version and when.
+
+This blog provides an example of how phased Strimzi upgrades can be performed.
+By using kustomize to create a deployment that allows multiple Strimzi versions to run concurrently and the `STRIMZI_CUSTOM_RESOURCE_SELECTOR` to allow us to specify which Strimzi instance manages which Kafka cluster.
+
+An [example repository](https://github.com/tomncooper/strimzi-phased-upgrade) is provided showing the final setup and providing example scripting and templates for adding new Strimzi versions.
+
+## Background
```
scholzj, line 19: If you move all headers one level down (i.e. `###` instead of `##`, `####` instead of `###`), the actual blog post on the web will look much better and the headers will fit with the rest of the website.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+First, lets go over what a Strimzi upgrade involves:
+
+### CRDs
+
+The first step in a Strimzi operator upgrade is the Custom Resource Definitions (CRDs) for the Strimzi Custom Resources (CRs): `Kafka`, `KafkaTopic`, `KafkaUser` etc. 
+These are Kubernetes cluster wide resources and need to match the most recent version of Strimzi you have installed on the Kubernetes Cluster. 
+They need to match the most recent version as new fields could have been added to the CRDs that the new Strimzi version expects to be there. 
+Strimzi CRDs are backwards compatible (within the same API version), so older Strimzi versions are ok with newer CRDs (they will ignore the new fields) but not the other way around.
+
+### Strimzi Deployment
```
scholzj, line 34: Should you also mention the cluster roles and cluster role bindings? They will also conflict between parallel operator installations so it should be worth mentioning them.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+However, when you have a large fleet of Kafka clusters being managed by Strimzi, there are several elements you must keep in mind when performing the upgrade. 
+
+First, lets go over what a Strimzi upgrade involves:
+
+### CRDs
+
+The first step in a Strimzi operator upgrade is the Custom Resource Definitions (CRDs) for the Strimzi Custom Resources (CRs): `Kafka`, `KafkaTopic`, `KafkaUser` etc. 
+These are Kubernetes cluster wide resources and need to match the most recent version of Strimzi you have installed on the Kubernetes Cluster. 
+They need to match the most recent version as new fields could have been added to the CRDs that the new Strimzi version expects to be there. 
+Strimzi CRDs are backwards compatible (within the same API version), so older Strimzi versions are ok with newer CRDs (they will ignore the new fields) but not the other way around.
```
scholzj, line 32: Nope, it is not so simple. For most versions, this is valid. But there are always versions when the CRD API versions are changing that will not work in parallel. For example from the past versions, Strimzi 0.21 and 0.23 cannot run in parallel as they will have different CRDs (due to `v1beta2`). Similar situations will likely arise in the future for `v1` API.

While today nobody knows how exactly will the versions look like for `v1` API, the only real guarantee is that you can run in parallel two Strimzi versions that follow each other. (E.g. in the old example you can run 0.21 with 0.22 and 0.22 with 0.23. But not 0.21 or earlier with 0.23 and later.)
The guarantee

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+So what can we do to add more control?
+
+## Phased upgrades
+
+What we need is a way to control which Kafka clusters get upgraded and in what order. 
+This is often referred to as a phased upgrade or as an upgrade with _"blast radius"_ control because we can limit the application of the upgrade to a small number of resources (the radius) and contain any potential negative effects (the blast).
+
+There are several possible ways to implement phased upgrades of the Kafka clusters managed by a Strimzi operator:
+
+### Custom image per Kafka cluster
```
scholzj, line 93: This would IMHO not work for most versions and has possibly many different fallouts. You should at least remove this section if not even mention it explicitly as something to absolutely avoid!

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+As you have manually set this, the changes to the image map environment variables when you upgrade the Strimzi Operator would not change the images of the operands.
+So no automatic rolling would occur.
+You could then manually switch the image used by each Kafka cluster to the new Strimzi version. 
+This would allow you to control which clusters rolled and when.
+
+However, you would then enter a situation where a new operator version (for example 0.45) is talking to a Kafka cluster running the old Strimzi version's container images (eg 0.44). 
+As mentioned in the introduction, the Strimzi container images represent an API with certain scripts, endpoints and environment variables expected to be present. 
+Running a newer Strimzi version administering a Kafka cluster using older Strimzi images _could_ lead to issues and is not a recommended or supported deployment model.
+
+### Strimzi per Namespace
```
scholzj, line 106: It might be worth to emphasize here that only Strimzi versions close to each other can share the CRDs, ClusterRoles etc., so while this is valid way, it requires a lot of self-control to make sure the versions do not diverge too much.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+The Cluster Operators would also all share the cluster-wide CRDs and so care and time would need to be taken to update these and upgrade all the Operators in a timely fashion.
+
+### Pause Reconciliation
+
+If you have a Strimzi Cluster Operator managing multiple Kafka clusters, either across a whole Kubernetes cluster or in a namespace, then another option is to temporarily stop the Operator from managing a given Kafka cluster (or other operand) until the upgrade is complete.
+
+To do this you would need to:
+1) Apply the [pause reconciliation annotation](https://strimzi.io/docs/operators/latest/deploying#proc-pausing-reconciliation-str) to all Kafka CRs (and other operands like Kafka Connect deployments)
+2) Upgrade the CRDs and the Strimzi Operator Deployment to the new version.
+3) In the order you want, remove the annotation from each Kafka CR. The operator will resume reconciliation, update the image reference and roll the Kafka cluster.
```
scholzj, line 123: This does not seem to render. Do you need to use `1.`, `2.` etc.? https://deploy-preview-478--strimzi.netlify.app/blog/2025/03/20/phased-strimzi-upgrade-example/

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+However, in order to use this process we need to deploy multiple versions of Strimzi concurrently.
+
+### Concurrent Strimzi Deployments
+
+The standard Strimzi deployment includes service accounts and RBAC resources (`Role`s, `ClusterRole`s and their associated bindings) needed for Strimzi to do its job. 
+By default these are all named `strimzi-cluster-operator` or similar.
+If we try and deploy multiple operator instances in the same namespace, or over multiple namespaces but needing to share `ClusterRole`s and bindings, things will get messy.
+Also, different Strimzi versions may need different permissions and so we need to make sure independent copies of these RBAC resources are deployed for each Strimzi version.
+
+#### Deploying Strimzi with Kustomize
```
scholzj, line 164: If you are happy with this, no need to change. But I wonder if it would be easier to just use different namespaces for each operator version?

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+
+```shell
+kubectl create --dry-run=client -o yaml \
+  clusterrolebinding strimzi-cluster-operator-namespaced \
+  --clusterrole=strimzi-cluster-operator-namespaced \
+  --serviceaccount strimzi:strimzi-cluster-operator \
+  > install-files/cluster-operator-0.43.0/strimzi-cluster-operator-namespaced.yaml
+```
+
+We can then add these additional resources to the `resources` list in the respective `kustomization.yaml` files and they will be modified by the same `nameSuffix` and `patch-role-reference.yaml` patches as the other RBAC resources.
```
scholzj, line 443: Doesn't this just create a new installation file while keeping the old RoleBinding in place as well?

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+```shell
+kubectl -n strimzi apply -k install-files
+```
+
+Note that the namespace you install the Operator set into needs to match the one you have used in the `patch-role-references.yaml`.
+
+A full example of the Operator Set above can be seen in the [strimzi-phased-upgrade](https://github.com/tomncooper/strimzi-phased-upgrade) repository.
+That repository also contains scripting to automate the addition of new Strimzi versions and templates for the generation of patch and kustomization files. 
+
+### The Phased Upgrade Process
```
scholzj, line 506: As I said above, I think you should maybe operate with just two versions as three might become too much.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+	style S_043 stroke:#f66,stroke-width:2px
+    S_044(Strimzi 0.44) --> K_A(Kafka A)
+	S_044 --> K_B(Kafka B)
+	S_045(Strimzi 0.45) --> K_C(Kafka C)
+```
+
+At this point the Strimzi 0.43 operator is not managing any Kafka clusters and all the clusters being managed by the 0.44 operator have been stable for sufficient period that we are confident we don't need to roll back. 
+We can now remove the Strimzi 0.43 operator deployment and its associated RBAC resources.
+
+### Things to watch out for
```
scholzj, line 609: You should add here also a warning about not removing the CRDs or ClusterRoles by mistake when deleting the old version. Especially with the CRDs, it would wipe all operands. This is the main issue of this approach.

scholzj (changes requested):
I think this is a good topic to cover in a blog post. I left some comments with some things I think it needs to cover.

_includes/site-scripts.html
tomncooper: I made it so the mermaid js is loaded based on a flag in the front matter.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: Oh I absolutely agree, I only mentioned it as I saw this proposal talked about seriously in the wild and wanted to explain why it was a bad idea.

But removing it seems reasonable, it is not a legitimate option and the post is already quite long.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+
+An [example repository](https://github.com/tomncooper/strimzi-phased-upgrade) is provided showing the final setup and providing example scripting and templates for adding new Strimzi versions.
+
+## Background
+
+Strimzi handles the [upgrade of Kafka clusters](https://strimzi.io/docs/operators/latest/deploying#proc-upgrade-kafka-kraft-str) for you, making it simple and easy for infrastructure teams to handle these complex operations. 
+The [upgrade of the Strimzi operator](https://strimzi.io/docs/operators/latest/deploying#assembly-upgrade-cluster-operator-str) itself is also straight-forward. 
+However, when you have a large fleet of Kafka clusters being managed by Strimzi, there are several elements you must keep in mind when performing the upgrade. 
+
+First, lets go over what a Strimzi upgrade involves:
```
ppatierno, line 25:
```suggestion
First, let's go over what a Strimzi upgrade involves:
```

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+### CRDs
+
+The first step in a Strimzi operator upgrade is the Custom Resource Definitions (CRDs) for the Strimzi Custom Resources (CRs): `Kafka`, `KafkaTopic`, `KafkaUser` etc. 
+These are Kubernetes cluster wide resources and need to match the most recent version of Strimzi you have installed on the Kubernetes Cluster. 
+They need to match the most recent version as new fields could have been added to the CRDs that the new Strimzi version expects to be there. 
+Strimzi CRDs are backwards compatible (within the same API version), so older Strimzi versions are ok with newer CRDs (they will ignore the new fields) but not the other way around.
+
+### Strimzi Deployment
+
+Once the CRDs are updated, you then need to [upgrade the Strimzi Operator](https://strimzi.io/docs/operators/latest/deploying#assembly-upgrade-cluster-operator-str) its self.
```
ppatierno, line 36:
```suggestion
Once the CRDs are updated, you then need to [upgrade the Strimzi Operator](https://strimzi.io/docs/operators/latest/deploying#assembly-upgrade-cluster-operator-str) itself. 
```

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+- 049-Crd-kafkarebalance.yaml
+- 04A-Crd-kafkanodepool.yaml
+```
+
+`install-files/crds/kustomization.yaml`
+```yaml
+resources:
+- cluster-operator-0.43.0
+- crds
+```
```
ppatierno, line 258: this long list of folders + files could easily confuse readers to understand what they are looking at. Maybe a short description for each would help?

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+- 010-ServiceAccount-strimzi-cluster-operator.yaml
+...
+- 060-Deployment-strimzi-cluster-operator.yaml
+
+nameSuffix: '-0-44-0'
+```
+
+#### Patching Environment Variables
+
+Sadly, the `nameSuffix` is not sufficient on it own.
```
ppatierno, line 318:
```suggestion
Sadly, the `nameSuffix` is not sufficient on its own. 
```

_posts/2025-03-20-phased-strimzi-upgrade-example.md
ppatierno: Yeah. Maybe it should also be mentioned at the beginning when dealing with pros and cons.

ppatierno (commented):
Great work! I had a first pass.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: So this section was supposed to be about what you need to do for a normal Stimzi upgrade and the consequences of that, basically setting up why you might want to have multiple strimzi versions installed.

The ClusterRole/Roles and their bindings are only an issue once you do that and we haven't proposed that yet at this point in the post.

scholzj: Yet you talk about these things with CRDs right above this, or?

fvaleri (commented):
Interesting post. One thing that is missing is how the user can check if the CRDs are compatible with all the operators versions. Probably there isn't an easy answer, but some recommendations would still be useful.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: This is a good point and why I put the comment in parentheses:

> Strimzi CRDs are backwards compatible (within the same API version)

However, you are right that this needs to be made far more explicit and the consequences discussed in the later text. 

Would it be too strong an assertion to say words to the effect of: 

> provided that the API version is the same, older versions _could_ work with newer CRDs. However the only guaranteed compatibility is between adjacent version, so testing is vital if you plan to have more than two concurrent versions.

I will also add the proviso that if you do have more than 2 concurrent versions and there is an CRD API version change, you will need to upgrade all Kafka clusters to the lower version (the one before the CRD API upgrade) first and then perform the CRD upgrade to the new version (no skipping of the transition Strimzi version).

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: It is definatly something to be mindful of so I will add it here. I'm not sure about adding at the beginning as it is a risk with a normal Strimzi deployment as well.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
scholzj: I think that as long as the API version doesn't change, we do provide backward compatibility. But the problem is that this alone does not provide a future-proof pattern. You essentially do not know exactly when the next API version will arrive. And if you end up having too many different Strimzi versions in use, they might work fine together, but you might find yourself suddenly unable to upgrade as the _next_ version would not work with them. This might be a nasty trap since you might be left out of bugfixes, CVE fixes etc. The ZooKeeper removal makes this even harder as the new versions with `v1` API won't support ZooKeeper anymore, so all Kafka clusters would need to be KRaft-based before you upgrade to v1 API.

With `v1`, we might have hopefully more time to support both versions in parallel as we will not be forced to move quickly by Kubernetes changes. But at this point, we do not know how many versions would be feasible.

So I think the important message here that we need to convey is that:
* These selectors should be fine to use when you want to do the phased upgrade from two adjacent Strimzi versions (which is the core of your blog post, but I suspect many readers will pick out of it what suites them)
* If you want to use this for a longer time to run multiple Strimzi versions in parallel, you need to be aware of the possible consequences. This is also the reason why we were rather silent on this feature (but I'm not saying we should not have this blog post about it - I think it is great to have it, we just need to cover the risks and explain them).

tomncooper: Totally agree, I will make the caveats much stronger and refactor the examples and repo around having just two versions.

_layouts/base.html
```diff
@@ -12,6 +12,10 @@
   {% include project-footer.html %}
 
   {% include site-scripts.html %}
+
+  {% if page.mermaid %}
```
scholzj, line 16: I guess you have to set this flag on your page now? (because right now it is not rendering)

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+You can prevent the fleet rolling situation by overriding the [image](https://strimzi.io/docs/operators/latest/configuring#con-common-configuration-images-reference) used by each Kafka cluster. 
+You could manually set this to the version used by the current Strimzi Operator version.
+As you have manually set this, the changes to the image map environment variables when you upgrade the Strimzi Operator would not change the images of the operands.
+So no automatic rolling would occur.
+You could then manually switch the image used by each Kafka cluster to the new Strimzi version. 
+This would allow you to control which clusters rolled and when.
+
+However, you would then enter a situation where a new operator version (for example 0.45) is talking to a Kafka cluster running the old Strimzi version's container images (eg 0.44). 
+As mentioned in the introduction, the Strimzi container images represent an API with certain scripts, endpoints and environment variables expected to be present. 
+Running a newer Strimzi version administering a Kafka cluster using older Strimzi images _could_ lead to issues and is not a recommended or supported deployment model.
```
scholzj, line 104: Worth adding to this - I forgot it in the initial comment - even with the same image, in most cases the Pods will still roll because there will be other changes as well. E.g. some new setting in the Pod, slightly different certificate subjects, configuration files etc. So even leaving aside the problems, it might still not help to avoid the RU.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
scholzj, line 123: You resolved the original comment, but it is still wrong.

_layouts/base.html
tomncooper: Yes, I haven't pushed the updated post markdown yet.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: Sorry, haven't actually pushed up the edited post markdown file yet. I should have done that before resolving.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: Yeah, it is a terrible idea. I have deleted the whole section (in the edited markdown I haven't pushed yet), so we don't accidentally encourage anyone into trying it anyway.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: Ok I have added a much stronger discussion of the CRD constraints to the start of the Resource Selector Upgrade Section. I have moved the examples to 2 versions of strimzi only and reiterated warnings in the summary and "Things to watch out for" section.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: Yes, but with this setup all the RBAC is completely seperate, different names pointing at different roles and different service accounts.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: I added this caveat to the summary section at the top of the post.

_posts/2025-03-20-phased-strimzi-upgrade-example.md
tomncooper: I added more descriptive captions for each listing, hopefully that makes it clearer what each relates too.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+
+### Background
+
+Strimzi handles the [upgrade of Kafka clusters](https://strimzi.io/docs/operators/latest/deploying#proc-upgrade-kafka-kraft-str) for you, making it simple and easy for infrastructure teams to handle these complex operations. 
+The [upgrade of the Strimzi operator](https://strimzi.io/docs/operators/latest/deploying#assembly-upgrade-cluster-operator-str) itself is also straight-forward. 
+However, when you have a large fleet of Kafka clusters being managed by Strimzi, there are several elements you must keep in mind when performing the upgrade. 
+
+#### Custom Resource Definitions 
+
+The first step in a Strimzi operator upgrade is the Custom Resource Definitions (CRDs) for the Strimzi Custom Resources (CRs): `Kafka`, `KafkaTopic`, `KafkaUser` etc (we refer to these as the Strimzi operator's _operands_).
```
fvaleri, line 57:
```suggestion
The first step in a Strimzi operator upgrade is to update the Custom Resource Definitions (CRDs) for the Strimzi Custom Resources (CRs): `Kafka`, `KafkaTopic`, `KafkaUser` etc (we refer to these as the Strimzi operator's _operands_). 
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+
+Using the Custom Resource Selector allows us to have fine grained control over which operands are upgraded to new Strimzi versions and when. 
+However, in order to use this process we need to deploy multiple versions of Strimzi concurrently.
+That requires us to consider several key aspects of the Strimzi installation resources:
+
+#### Concurrent Strimzi Deployments
+
+##### Custom Resource Definitions
+
+The first major consideration when designing a deployment strategy for multiple Strimzi versions is the [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions) (CRDs).
```
fvaleri, line 177: You already introduced the CRDs in a previous paragraph, so you can link the doc there, and there is no need to define the acronym again.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+For example operands being managed by Strimzi 0.43 when the 0.45 CRDs are installed in the cluster.
+This is an untested configuration and so you may encounter issues.
+Therefore, it is strongly recommended that you only run two adjacent Strimzi versions using a system like the one proposed in this post.
+
+###### CRD API version upgrades
+
+The next consideration is when the CRD API version is upgraded.
+For example when Strimzi eventually upgrades from `v1beta2` to `v1` CRDs.
+When this occurs you will need to first make sure **all** operands are updated to the latest Strimzi version running the old CRD API version (e.g. `v1beta1`).
+You will need to do this **before** you install the CRDs for the new Strimzi version which uses the new CRD API version (e.g. `v1`)
```
fvaleri, line 201:
```suggestion
You will need to do this **before** you install the CRDs for the new Strimzi version which uses the new CRD API version (e.g. `v1`).
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+
+nameSuffix: '-0-44-0'
+```
+
+##### Patching Environment Variables
+
+Sadly, the `nameSuffix` is not sufficient on its own. 
+There are several other elements of the deployment files that need to be customised in order to allow the Strimzi versions to exist concurrently.
+
+The first of these are the environment variables in the Strimzi operator `Deployment` resource, several of which need to be individualised to each Strimzi version.
```
PaulRMellor, line 376: Not sure about this _individualised_ in the doc
I might go for something like tailoring or customizing

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+patches:
+  - path: patches/patch-deployment.yaml
+    target:
+      kind: Deployment
+      name: strimzi-cluster-operator
+```
+
+##### Patching Role and ClusterRole Bindings
+
+Another element that needs individualising for each Strimzi version are the `RoleBinding`s and `ClusterRoleBinding`s used by the operator.
```
PaulRMellor, line 420:
```suggestion
The `RoleBinding` and `ClusterRoleBinding` resources used by the operator also need to be tailored for each Strimzi version.
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+
+patches:
+  ...
+  - path: patches/patch-role-references.yaml
+    target:
+      kind: (RoleBinding|ClusterRoleBinding)
+      name: ".*"
+```
+
+This patch will, for _all_ `RoleBinding` and `ClusterRoleBinding` in the `resources` list, replace the name of the first entry in the `subjects` list (in the base installation files this always the operator `ServiceAccount`) with the name specified in the patch.
```
PaulRMellor, line 450:
```suggestion
This patch updates all `RoleBinding` and `ClusterRoleBinding` resources in the resources list. 
It replaces the name of the first entry in the subjects list (which is the operator’s `ServiceAccount` in the base installation files) with the name specified in the patch.
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+It will then roll the cluster, one broker at a time, on to the new image.
+
+```mermaid
+flowchart TD
+  S_043(Strimzi 0.43) --> K_A(Kafka A)
+  S_043 --> K_B(Kafka B)
+	S_044(Strimzi 0.44) --> K_C(Kafka C)
+```
+
+After a burn-in period, where we are happy with the stability of the cluster, our client side canary jobs are stable and no client teams are reporting issues, we begin to move the rest of the fleet.
```
PaulRMellor, line 616: "our client side canary jobs"  comes a bit out-of-the-blue
Will it raise any questions in the reader?
Could we keep this neutral " where the cluster remains stable and no issues are reported"

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+```mermaid
+flowchart TD
+  S_043(Strimzi 0.43)
+	style S_043 stroke:#f66,stroke-width:2px
+  S_044(Strimzi 0.44) --> K_A(Kafka A)
+  S_044 --> K_B(Kafka B)
+	S_044 --> K_C(Kafka C)
+```
+
+At this point the Strimzi 0.43 operator is not managing any Kafka clusters and all the clusters being managed by the 0.44 operator have been stable for sufficient period that we are confident we don't need to roll back.
```
PaulRMellor, line 654:
```suggestion
At this point the Strimzi 0.43 operator is not managing any Kafka clusters and all the clusters being managed by the 0.44 operator have been stable for a sufficient period that we are confident we don't need to roll back. 
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+  S_044 --> K_B(Kafka B)
+	S_044 --> K_C(Kafka C)
+  S_045(Strimzi 0.45)
+	style S_045 stroke:#008000,stroke-width:2px
+```
+
+#### Things to watch out for
+
+The main things to watch for when using this Strimzi deployment pattern are:
+- Take note of the advice in the [CRD section](#custom-resource-definitions-1) above around the number of concurrent versions and CRD API version upgrades:
```
PaulRMellor, line 670:
```suggestion
- Take note of the advice in the [CRD section](#custom-resource-definitions-1) around the number of concurrent versions and CRD API version upgrades:
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+  - The only supported configuration is running two adjacent Strimzi versions concurrently.
+  - When upgrading the CRD API version. You will need to upgrade all operands to the last Strimzi version before the API upgrade, before adding the new Strimzi version with the new CRD API version.
```
PaulRMellor, line 672:
```suggestion
  - When upgrading the CRD API version, you must first upgrade all operands to the last Strimzi version that uses the old API, before adding the new Strimzi version with the updated CRDs.
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+- Extreme care should be taken, while updating the Operator Set, to not remove the CRDs from the Kubernetes cluster. Doing so would result in the removal of **all** Kafka clusters (and other Strimzi operands).
```
PaulRMellor, line 673:
```suggestion
- Take extreme care when updating the Operator Set to avoid removing the CRDs from the Kubernetes cluster. Doing so would delete all Kafka clusters and other Strimzi-managed resources.
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+- You should ensure that the Kafka version of the Kafka CR is supported by the version of Strimzi you are moving too. This should not be an issue of are moving between adjacent Strimzi versions. However, you can add checks to deployment scripts and GitOps repository CI that checks for this.
```
PaulRMellor, line 674:
```suggestion
- You should ensure that the Kafka version of the Kafka CR is supported by the version of Strimzi you are moving too. This should not be an issue if you are moving between adjacent Strimzi versions. However, you can add checks to deployment scripts and GitOps repository CI.
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+
+#### Further deployment options
+
+We have covered how to do upgrades using the custom resource selector and multiple different versions of Strimzi. 
+However, there is no reason there should be only one instance of each version.
+You could have `strimzi-0-43-0-a` and `stimzi-0-43-0-b`, you could then load balance Kafka cluster management between them.
+This would be particularly useful if you have a large number of small Kafka clusters or _very_ large Kafka clusters with many brokers.
+
+You could also separate out critical Kafka cluster management to a Strimzi operator with different RBAC.
+Perhaps your infrastructure management system itself runs though a Kafka cluster managed by Strimzi, you could have a separate Strimzi operator deployed just to manage this critical cluster and lock it down with very restrictive controls.
```
PaulRMellor, line 685:
```suggestion
Perhaps your infrastructure management system itself runs through a Kafka cluster managed by Strimzi. 
In that case, you could deploy a separate Strimzi operator dedicated to managing this critical cluster and lock it down with restrictive controls.
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,700 @@
...
+---
+
+### Summary
+
+When you upgrade an instance of the Strimzi Cluster Operator you will force **all** the Kafka clusters that it manages to roll. 
+For large fleets this could be quite disruptive.
+It would be nice to be able to decide which Kafka clusters upgrade to the new Strimzi version and when.
+
+This post provides an example of how phased Strimzi upgrades can be performed.
+By using [kustomize](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/) to create a deployment that allows multiple Strimzi versions to run concurrently and the `STRIMZI_CUSTOM_RESOURCE_SELECTOR` environment variable to allow us to specify which Strimzi instance manages which Kafka cluster via labels on the Kafka custom resources.
```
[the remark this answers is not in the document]
PaulRMellor: Is the official name uppercase Kustomize?

PaulRMellor (approved):
Great post, Tom — really well explained and easy to follow.
I’ve added a few suggestions for readability and flow as I went through.

One thing to consider: the CRD/API compatibility advice is repeated in several places. It might be a bit too detailed for the intro.
Would it make sense to consolidate that guidance under the CRD API version upgrades and Supported concurrent versions sections, and refer back to those from the intro and the "Things to watch out for" section?

_posts/2025-03-20-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,637 @@
...
+              3.8.0=quay.io/strimzi/kafka:0.45.0-kafka-3.8.0
+              3.8.1=quay.io/strimzi/kafka:0.45.0-kafka-3.8.1
+              3.9.0=quay.io/strimzi/kafka:0.45.0-kafka-3.9.0
+        ...
+```
+
+In the sample above you can see that that Kafka version maps to an image which contains the Strimzi version. 
+The container images used by Strimzi include scripts, health endpoint servers and other services which form an API for the container.
+A given version of Strimzi expects certain endpoints, scripts etc to be present on the container in order to perform its operations. 
+This is why, even if the Kafka version has not changed, the image must change when you upgrade Strimzi.
```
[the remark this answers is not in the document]
tomncooper: That is a good point, but I feel like it is orthogonal to the point of this post. However, we should definitely highlight the custom image Strimzi upgrade issue in the docs (if we don't already).

_posts/2025-03-26-phased-strimzi-upgrade-example.md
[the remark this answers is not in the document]
tomncooper: From reviewing their documentation, it appears not.

fvaleri (approved):
LGTM. Thanks for addressing my comments.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
tomncooper: I agree, _tailored_ feels cleaner.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
tomncooper: Yes, that is a bit out of the blue. I will simplify it.

tomncooper:
> Would it make sense to consolidate that guidance under the CRD API version upgrades and Supported concurrent versions sections, and refer back to those from the intro and the "Things to watch out for" section?

Thanks for the review @PaulRMellor. I think you are right, I need to rationalise the CRD advice, it is in the wrong place and messes up the flow a bit. I will have a go at restructuring.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,702 @@
...
+
+### Background
+
+Strimzi handles the [upgrade of Kafka clusters](https://strimzi.io/docs/operators/latest/deploying#proc-upgrade-kafka-kraft-str) for you, making it simple and easy for infrastructure teams to handle these complex operations. 
+The [upgrade of the Strimzi operator](https://strimzi.io/docs/operators/latest/deploying#assembly-upgrade-cluster-operator-str) itself is also straight-forward. 
+However, when you have a large fleet of Kafka clusters being managed by Strimzi, there are several elements you must keep in mind when performing the upgrade. 
+
+#### Custom Resource Definitions 
+
+The first step in a Strimzi operator upgrade is to update the [Custom Resource Definitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions) (CRDs) for the Strimzi Custom Resources (CRs): `Kafka`, `KafkaTopic`, `KafkaUser` etc (we refer to these as the Strimzi operator's _operands_).
```
scholzj, line 57: For me, the operand is the running _component_ operated by Strimzi. Not sure I would call the CR _operand_ - even double so for topics and users which do not have the traditional style _operand_ at all.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,702 @@
...
+The CRDs are Kubernetes cluster-wide resources and need to match the most recent version of Strimzi you have installed on the Kubernetes cluster. 
+This is because new fields could have been added to the CRDs that the new Strimzi version expects to be there. 
+Strimzi CRDs are backwards compatible (within the same API version -- see [this section](#custom-resource-definitions-1) for more details), so older Strimzi versions are ok with newer CRDs (they will ignore the new fields) but not the other way around.
+
+#### Strimzi Deployment
+
+Once the CRDs are updated, you then need to [upgrade the Strimzi Operator](https://strimzi.io/docs/operators/latest/deploying#assembly-upgrade-cluster-operator-str) itself. 
+This requires you to update the [Deployment CR](https://github.com/strimzi/strimzi-kafka-operator/blob/main/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml) (and any additional RBAC or other resources the new version needs).
```
scholzj, line 65:
```suggestion
This requires you to update the [Deployment resource](https://github.com/strimzi/strimzi-kafka-operator/blob/main/install/cluster-operator/060-Deployment-strimzi-cluster-operator.yaml) (and any additional RBAC or other resources the new version needs). 
```

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,702 @@
...
+
+```mermaid
+flowchart TD
+  S_044(Strimzi 0.44) --> K_A(Kafka A)
+  S_044 --> K_B(Kafka B)
+	S_044 --> K_C(Kafka C)
+  S_045(Strimzi 0.45)
+	style S_045 stroke:#008000,stroke-width:2px
+```
+
```
scholzj, line 667: Maybe you can here at the end mention that the same process can be used for Connect, MM2, and Bridge?

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,702 @@
...
+- Take extreme care when updating the Operator Set to avoid removing the CRDs from the Kubernetes cluster. Doing so would delete all Kafka clusters and other Strimzi-managed resources.
+- You should ensure that the Kafka version of the Kafka CR is supported by the version of Strimzi you are moving too. This should not be an issue if you are moving between adjacent Strimzi versions. However, you can add checks to deployment scripts and GitOps repository CI.
+- When removing old versions of Strimzi that are no longer needed, care should be taken to make sure that no Kafka CRs are still labelled for that version. Again, scripting or CI can run checks against the Kubernetes cluster to ensure this.
+
+#### Further deployment options
+
+We have covered how to do upgrades using the custom resource selector and multiple different versions of Strimzi. 
+However, there is no reason there should be only one instance of each version.
+You could have `strimzi-0-43-0-a` and `stimzi-0-43-0-b`, you could then load balance Kafka cluster management between them.
```
scholzj, line 682: I think _load balancing_ suggests it happens automatically. Would sharding be better here?

scholzj (commented):
Some mroe nits and one important point about the pausing of reconciliation.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,702 @@
...
+When this occurs you will need to first make sure **all** operands are updated to the latest Strimzi version running the old CRD API version (e.g. `v1beta1`).
+You will need to do this **before** you install the CRDs for the new Strimzi version which uses the new CRD API version (e.g. `v1`).
+You would then follow any specific upgrade instructions and move each operand onto the new Strimzi version and CRD API version.
+
+Using the last CRD API version upgrade as an example, we can illustrate the process. Strimzi 0.22 uses `v1beta1` and Strimzi 0.23 uses `v1beta2` CRDs. If we were running a setup with Strimzi 0.21 and 0.22 installed concurrently, to move to 0.23 we would need to:
+- First move **all** Kafka clusters, Kafka Connect deployments and other operands to be run by the Strimzi 0.22 Operator.
+- Then remove Strimzi 0.21 and only then add Strimzi 0.23 and its CRDs. This step is important as 0.21 based operands **cannot** run with 0.23 CRDs. 
+- Finally move every operand across to the 0.23 Operator and perform the specific [CRD upgrade instructions](https://strimzi.io/docs/operators/0.23.0/deploying#assembly-upgrade-resources-str).
+- Remove the 0.22 Operator.
```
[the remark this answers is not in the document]
tomncooper: Your suggestion is clearer. I have tweaked it a bit but the meaning is what you proposed.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,702 @@
...
+```shell
+kubectl -n strimzi apply -k install-files
+```
+
+Note that the namespace you install the Operator Set into needs to match the one you have used in the `patch-role-references.yaml`.
+
+A full example of the Operator Set above can be seen in the [strimzi-phased-upgrade](https://github.com/tomncooper/strimzi-phased-upgrade) repository.
+That repository also contains scripting to automate the addition of new Strimzi versions and templates for the generation of patch and kustomization files. 
+
+#### The Phased Upgrade Process
```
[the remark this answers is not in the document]
tomncooper: Ahh that is a very good point. I have updated the walk-through.

_posts/2025-03-26-phased-strimzi-upgrade-example.md
tomncooper: Good point, I went with "distribute between them".

_posts/2025-04-09-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,733 @@
...
+So, if we were running a setup with Strimzi 0.21 and wanted to upgrade to 0.23, we have to:
+
+1. First install CRDs from Strimzi 0.22 with both `v1beta1` and `v1beta2` API versions and install the Stirmzi 0.22 deployment, RBAC resources, etc.
+1. Migrate all Strimzi managed CRs from 0.21 to 0.22.
+1. Remove the 0.21 operator.
+1. Perform the specific [CRD upgrade instructions](https://strimzi.io/docs/operators/0.23.0/deploying#assembly-upgrade-resources-str), install CRDs from Strimzi 0.23 with only the `v1beta2` API versions and install the Stirmzi 0.23 deployment, RBAC resources, etc.
+1. Migrate all Strimzi managed CRs from 0.22 to 0.23.
+1. Remove the 0.22 operator.
+
+##### Strimzi install resources
```
scholzj, line 223: `Strimzi install resources` sounds strange to me. Feel free to keep it if it is intentional and makes sense to you. But maybe `Strimzi installation resources` would be better?

_posts/2025-04-09-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,733 @@
...
+  name: strimzi-cluster-operator
+spec:
+  template:
+    spec:
+      containers:
+        - name: strimzi-cluster-operator
+          env:
+            - name: STRIMZI_NAMESPACE
+              value: "*" 
+              valueFrom: null
```
scholzj, line 494: Is the `valueFrom: null` intentional here? I assume it is valid Deployment this way ... but it looks strange and would be normally omitted.

_posts/2025-04-09-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,733 @@
...
+We now want to upgrade to a newer Strimzi version. 
+The Kafka C cluster is our most risk tolerant cluster, it hosts topics used by low tier data pipelines which can tolerate extended downtime, so it will go first to ensure there are no issues.
+We first check that the Kafka C cluster is error free and in the `Ready` state. 
+We then apply the [pause reconciliation annotation](https://strimzi.io/docs/operators/latest/deploying#proc-pausing-reconciliation-str) to the Kafka CR so that the Strimzi operator currently managing it does not attempt to start an operation when we are about to move the CR:
+
+```shell
+kubectl annotate kafka kafka-c strimzi.io/pause-reconciliation="true"
+```
+
+We then check to make sure that reconciliation is actually paused by making sure the `ReconciliationPaused` condition is present on the Kafka CR.
```
scholzj, line 622: Do you want to include a command here for this? Something like (not tested)
```
kubectl wait kafka/my-cluster --for=condition=ReconciliationPaused --timeout=300s
```

_posts/2025-04-09-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,733 @@
...
+We then apply the Strimzi custom resource selector label for the 0.44.0 Strimzi operator:
+
+```shell
+kubectl -n kafka label --overwrite kafka kafka-c strimzi-resource-selector=strimzi-0-44-0
+```
+
+We can then unpause the reconciliation of the Kafka CR:
+
+```shell
+kubectl annotate kafka kafka-c strimzi.io/pause-reconciliation="true"
```
scholzj, line 632:
```suggestion
kubectl annotate kafka kafka-c strimzi.io/pause-reconciliation="false"
```

scholzj (approved):
Few more nits.But LGTM otherwise. No need to re-review.

_posts/2025-04-09-phased-strimzi-upgrade-example.md
tomncooper: Yes it does sound wrong, particularly because everywhere else in the doc apart from this title I have called it "Strimzi installation resources"?

_posts/2025-04-09-phased-strimzi-upgrade-example.md
tomncooper: So this is required in order to negate the existing `valueFrom` in the base resource the patch is applied to. The resulting output would not contain the `valueFrom` field.

_posts/2025-04-09-phased-strimzi-upgrade-example.md
tomncooper: Well, that is embarrassing...

ppatierno (approved):
LGTM. Great article. I left just a couple of questions and advice but no need to review again.

_posts/2025-04-09-phased-strimzi-upgrade-example.md
```diff
@@ -0,0 +1,733 @@
...
+Leaving aside whether this is a good idea or not, using the setup proposed in this post, you could create an isolated operator instance just for that team which would not effect any other running Kafka clusters.
+
+### Conclusion
+
+In this post we have covered an example Strimzi deployment setup that allows you to deploy multiple versions of the Strimzi operator concurrently and pass Kafka cluster management between them on a cluster-by-cluster basis. 
+This allows you to control what Kafka clusters are upgraded and when and also allows easy rollback in the case of issues.
+
+The example [strimzi-phased-upgrade](https://github.com/tomncooper/strimzi-phased-upgrade) repository uses Kustomize to modify the installation resources of each Strimzi version to make sure they do not interfere with one another. 
+It also provides scripting to automate the addition of new Strimzi versions.
+However, a similar outcome could be achieved using [Helm](https://helm.sh/) or the [Operator Lifecycle Manager (OLM)](https://olm.operatorframework.io/) tool from the Operator Framework.
```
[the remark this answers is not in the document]
tomncooper: A part 2 with an OLM and/or Helm version would be interesting though.

tomncooper:
Ok, I think this is ready to publish if you are happy @ppatierno?

_posts/2025-04-09-phased-strimzi-upgrade-example.md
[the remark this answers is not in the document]
ppatierno: > Well you would need a bit more involved scripting to generate the bundle image containing the versions.

Yeah exactly, this is what I see as "problematic" or anyway not that simple. You cannot use the CSV bundle image out of box and as it is. I think it deserves a part 2 :-)

ppatierno:
> Ok, I think this is ready to publish if you are happy @ppatierno?

Absolutely ;-)

PR #509, opened 2025-06-24
Blog post about Kafka Queue

tinaselenge:
### Type of change

* [ ] Typo/minor fix
* [ ] Other


_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+
+### Comparing share and consumer groups
+
+I have set up a single node cluster on my machine using Strimzi’s [Quickstart](https://strimzi.io/quickstarts/) and created a topic called <code>kafka-queue</code> with 2 partitions. 
+
+I also added the following 2 configurations to my Kafka CR to enable this feature, as it is not enabled by default since it is not production ready yet:
+```
+group.coordinator.rebalance.protocols: classic,consumer,share
+unstable.api.versions.enable: true
+```
```
scholzj, line 135: It would be good to share the Kafka CR with these values - for examle in simplified format, e.g.:

```yaml
apiVersion: kafka.strimzi.io/v1beta2
kind: Kafka
metadata:
  # ...
spec:
  kafka:
    # ...
    config:
      group.coordinator.rebalance.protocols: classic,consumer,share
      unstable.api.versions.enable: true
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+
+In the traditional message queue, once a record is fetched by a consumer, it is no longer in the queue. You can think of <b>Archived</b> state, as the records that are taken off the queue and no longer available. 
+
+### Share Group State
+
+The share group persists state management for subscribed partitions, by storing it in an internal topic called, <code>__share_group_state</code>. This topic stores state of share groups, for example, the partitions they are subscribed to and the state of those partitions such as their fetched start offset (SPSO) and the end offset (SPEO) as well as the delivery state and delivery count of the records on the partitions.
+
+We can take a look at how this information is stored in the <code>__share_group_state</code> topic by using the console consumer tool:
+```
+./bin/kafka-console-consumer.sh --bootstrap-server localhost:9092 --topic __shared_group_state | jq .
```
scholzj, line 54: This seems a bit disconnected as users cannot run this with Strimzi. I wonder if it would be possible to move it only behind the chapter where you have the cluster and change the command to allow users to try this themself.

I also wonder how did the shared group got there -> wouldn't it show up there only after being used?

assets/images/posts/2025-06-24-kafka-queue-08.png
scholzj, line 1: This screenshot (as well as the second one) seems a bit strange, as you see the closing `}` but not the initial one, so it is kind of starting in the middle. I wonder if we can include the start of some method or something there?

Also, the screenshots are not accessible - it might be great to have also an alternative link to the code in GitHub Gist or something like that.

assets/images/posts/2025-06-24-kafka-queue-04.png
scholzj, line 1: Probably not worth changing the screenshots now ... but Ideally they should avoid execing into the broker pods and running the Java commands there as that can cause problems such as running out of memory etc.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
+---
+layout: post
+title:  "Using Kafka Queue feature with Strimzi"
+date: 2025-06-24
+author: tina_selenge
+---
+
+The Kafka Queue feature introduced by [KIP-932](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka) as an early access in Kafka version 4.0 which is supported by Strimzi 0.46.  In this blog, I’m going to introduce this feature and how you can try it out with your Strimzi managed cluster.
```
scholzj, line 8: It might be good to mention the early access limitations? IIRC it has also some upgrade impact which definitely should be mentioned with a warning to not use this in production?

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+| :--: |:-------------|
+| 0    | Available    |
+| 2    | Acknowledged |
+| 4    | Archived     |
+
+Let’s take a look at the following example of a partition that a share group is subscribed to, which is called <b>Share Partition</b>:
+
+![Delivery state](/assets/images/posts/2025-06-24-kafka-queue-03.png)
+
+In this example, offset 2 is the start offset that a share group is consuming from the partition and offset 2 and 4 are currently acquired by a consumer of the group for the first time, therefore delivery count is 1.  However, the record at offset 3 is in <b>Available</b> state with delivery count of 2, which means a consumer of the group has attempted to deliver this record twice. This record will be retried until the maximum delivery count is reached. The record at offset 5 has been processed successfully and acknowledged and the record at offset 6 is the next available record to be acquired by a consumer of the share group, therefore it is the end offset for this share partition.
```
showuon, line 125: This sentence is hard to read:
> In this example, offset 2 is the start offset that a share group is consuming from the partition and offset 2 and 4 are currently acquired by a consumer of the group for the first time, therefore delivery count is 1.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+kafka-consumer-2                              1/1     Running   0          2m34s
+kafka-share-consumer-0                        1/1     Running   0          2m9s
+kafka-share-consumer-1                        1/1     Running   0          2m
+kafka-share-consumer-2                        1/1     Running   0          110s
+my-cluster-dual-role-0                        1/1     Running   0          7m44s
+my-cluster-entity-operator-78d5b9dfd9-5ktxt   2/2     Running   0          10m
+strimzi-cluster-operator-5dd46b9985-dp2wt     1/1     Running   0          10m
+```
+
+The following screenshot of the split screen shows that I am sending records to the topic with 2 partitions (in left bottom terminal) and the other terminal windows are where the 3 share consumers are running printing the records with their partition number and offset.
```
showuon, line 164: `--record-size 1024` makes the terminal output hard to read. We can use size 10, or 3, ... to avoid to distract users.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+
+![KafkaShareConsumer API](/assets/images/posts/2025-06-24-kafka-queue-08.png)
+
+<i>Acknowledging individual records</i>
+
+This allows users to acknowledge an individual record depending on the outcome of the processing of the record. Each record is acknowledged using a call to <code>acknowledge()</code> which takes different types of acknowledgement as an argument: ACCEPT, RELEASE and REJECT.
+This aligns with the actions that can be taken by share group consumers mentioned previously.
+
+In the next example, I have used the new consumer configuration added for this feature, <b>share.acknowledgement.mode</b> and set it to "explicit". This configuration is set to "implicit" by default, which is why I didn’t need to set this configuration for the previous example.
```
showuon, line 199: >  which is why I didn’t need to set this configuration for the previous example. 

But I still see you set `share.acknowledgement.mode=explicit` in the previous example?

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+
+If the client encountered an error not caused by the JSON mapping, it releases the record for another attempt because it could be a transient failure.
+
+When no exception occurred during the processing, the consumer client accepts the record. Once all the records in the batch are processed and acknowledged individually, these states are stored locally in the consumer. Then the client has to call <code>commitSync()</code> or <code>commitAsync()</code> to commit the state to the internal topic, <code>__shared_group_state</code>. If any of the records in the batch hits an error or skips the processing, the client commits the state at that point. The records in the batch that were not processed or acknowledged yet, will be presented to the consumer client again as part of the same acquisition, therefore their delivery count will not be incremented.
```
showuon, line 209: > The records in the batch that were not processed or acknowledged yet, will be presented to the consumer client again as part of the same acquisition, therefore their delivery count will not be incremented.

I'm not familiar with this, but form the example below, the delivery count of offset 366 is incremented and then after 4th retried, it is archived. So it looks like it contradicts with each, right?

_posts/2025-06-24-kafka-queue.md
katheris: I think it might be nice to have the instructions here more directed at the reader, e.g. deploy the quickstart, then edit your Kafka CR to add these two settings...

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+
+<i>Inspecting __share_group_state topic</i>
```
katheris, line 211: Should this be a heading? Looking at the [preview](https://deploy-preview-509--strimzi.netlify.app/blog/2025/06/24/kafka-queue/) it's getting a little lost

katheris (commented):
I've added a few different comments about wording and things.

In terms of the overall structure. I wonder whether it might be nice to have the steps showing how to use the share group and comparing to a regular consumer group earlier, so e.g.

- High level overview of queues in Kafka
- Example of how to use both consumer group types and how they work different
- More in-depth discussion about how it works (the sections you have on partition assignment, fetch mechanism, delivery state
- Example of using the API

It might help the reader to understand better the difference before diving into some of the more technical details

@tinaselenge @scholzj @showuon what do you think?

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
+---
+layout: post
+title:  "Using Kafka Queue feature with Strimzi"
+date: 2025-06-24
+author: tina_selenge
+---
+
+The Kafka Queue feature introduced by [KIP-932](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka) as an early access in Kafka version 4.0 which is supported by Strimzi 0.46.  In this blog, I’m going to introduce this feature and how you can try it out with your Strimzi managed cluster. 
+
+This feature is only supported with KRaft clusters since Zookeeper was removed in the Kafka 4.0 release. It is also based on the new consumer rebalance protocol introduced by [KIP-848](https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol) that enhances stability, scalability and simplifies client implementations.
```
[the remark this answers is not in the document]
tinaselenge: The KIP moved partitions assignments from client to server, so that's what I meant by simplifying the client but this line is actually not clear, so I will just change it as you suggested with something more generic.

showuon:
> I've added a few different comments about wording and things.
> 
> In terms of the overall structure. I wonder whether it might be nice to have the steps showing how to use the share group and comparing to a regular consumer group earlier, so e.g.
> 
>     * High level overview of queues in Kafka
> 
>     * Example of how to use both consumer group types and how they work different
> 
>     * More in-depth discussion about how it works (the sections you have on partition assignment, fetch mechanism, delivery state
> 
>     * Example of using the API
> 
> 
> It might help the reader to understand better the difference before diving into some of the more technical details
> 
> @tinaselenge @scholzj @showuon what do you think?

I like this idea. Thanks Kate!

_posts/2025-06-24-kafka-queue.md
scholzj: TBH, I would not hold it - or to be clear, I do not have problem if others want to wait with it, but I see no need for it. Just make it more clear what the state and the plan is.

Even if it is production ready in 4.1 and we wait for Strimzi 0.48, we would still need to mention it as the blog post is not directly related to any particular release. So one way or another, this will need to be covered even if we wait.

assets/images/posts/2025-06-24-kafka-queue-08.png
tinaselenge: Instead of a screenshot, I put the example in a code block that can be copied.

assets/images/posts/2025-06-24-kafka-queue-04.png
tinaselenge: I might change this later on to run a container instead when I'm testing 4.1 version.

_posts/2025-06-24-kafka-queue.md
tinaselenge: I have left it for now, as I would need to try it out once 4.1 is released and is supported in Strimzi. If that happens before posting this blog (which might be the case anyway), I will update it.

_posts/2025-06-24-kafka-queue.md
tinaselenge: That is a mistake, thanks for catching that. I have removed it now.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,416 @@
...
+
+The Kafka Queue feature allows you to use Kafka like a traditional message queue and increase parallelism of message processing beyond the number of partitions. It provides queue-like semantics through a new consumer group called <b>shared group</b>. This new type of group gives more fine-grained control on message acknowledgement and retries. 
+
+The key difference between share group and regular consumer group is how partitions get assigned to consumer members. With regular consumer groups, each partition is exclusively assigned to a single member of the consumer group, therefore users typically have as many consumer members as the number of partitions to maximise the parallelism in message processing. However, share groups balance partitions between all members of a share group, allowing multiple consumer members to fetch from the same partition. So users can have more consumers than the number of partitions, further increasing the parallelism.
```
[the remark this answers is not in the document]
tinaselenge: good idea, I have included this now.

tinaselenge:
@katheris thank you so much for reviewing the blog again. I believe I addressed the recent comments, please let me know if I missed anything.

katheris (approved):
Thanks @tinaselenge, this looks great

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+
+The share group currently does not provide ordering guarantee as multiple consumers can fetch records from the same partition. And it does not support dead letter queue to handle unprocessable/poison records. However, these features may potentially be added in the future.
```
showuon, line 20: Should we also mention the share group doesn't support exactly-once semantics now given we've mentioned it above? In the [KIP](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka-FutureWork), it said:

> Finally, this KIP does not include support for acknowledging delivery using transactions for exactly-once semantics.

showuon (commented):
Not sure if I can review this now or later? @tinaselenge , Just want to confirm, are you waiting for the v4.1.0 release and want to update the post then?

tinaselenge:
> Not sure if I can review this now or later? @tinaselenge , Just want to confirm, are you waiting for the v4.1.0 release and want to update the post then?

If people are happy with it as it is, I'm happy to get it posted and then update later once we have v4.1.0 release.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+
+Let’s now take a look at the outcome out of this client:
+
+![KafkaShareConsumer API example](/assets/images/posts/2025-06-24-kafka-queue-08.png)
+
+The record at offset 361 was rejected because it hit an error mapping the record value into JSON therefore is considered a poison message. 
+The records at offset 362, 363, 364 and 365 were successfully processed and therefore accepted.
+The record at offset 366 was however released without getting processed therefore it was retried 5 times before getting archived.
+
+Let’s now inspect the state of these records from the internal topic, __share_group_state:
```
mimaison, line 389: Inspecting the data of an internal topic is really not the way we should tell people to monitor their share groups. I'm worried we keep showing this and it may end up looking like good practice.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+
+#4. Once a poison record is sent to the dead letter queue topic, the consumer client rejects it so that it gets archived immediately. 
+
+#5. After rejecting a bad record, it exits the loop for processing each record.
+
+#6. If the client encountered an error not caused by the JSON mapping, it releases the record for another attempt because it could be a transient failure.
+
+#7. When no exception occurred during the processing, the consumer client accepts the record. Once all the records in the batch are processed and acknowledged individually, these states are stored locally in the consumer. 
+
+#8. Then the client has to call <code>commitSync()</code> or <code>commitAsync()</code> to commit the state to the internal topic, <code>__share_group_state</code>. In #2 and #3, it exits the loop for processing each record, and commits the state by calling <code>commitSync()</code>. The records in the batch that were not processed or acknowledged yet, will be presented to the consumer client again as part of the same acquisition, therefore their delivery count will not be incremented.
```
mimaison, line 377: Genuine question as I don't know the answer, can you use auto commit with a share group? or do you have to explicitly commit yourself all the time?

mimaison (commented):
Thanks for the post! It's nice to have an overview of a new feature. I left a few comments and suggestions.

_posts/2025-06-24-kafka-queue.md
tinaselenge: I've changed the record size

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+kafka-consumer-2                              1/1     Running   0          2m34s
+kafka-share-consumer-0                        1/1     Running   0          2m9s
+kafka-share-consumer-1                        1/1     Running   0          2m
+kafka-share-consumer-2                        1/1     Running   0          110s
+my-cluster-dual-role-0                        1/1     Running   0          7m44s
+my-cluster-entity-operator-78d5b9dfd9-5ktxt   2/2     Running   0          10m
+strimzi-cluster-operator-5dd46b9985-dp2wt     1/1     Running   0          10m
+```
+
+The following screenshot of the terminal tabs shows that I am sending records to the topic with 2 partitions (in left bottom terminal) and the other terminal tabs are where the 3 share consumers are running, printing the records with their partition number and offset.
```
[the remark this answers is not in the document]
tinaselenge: I've updated the screenshot to show the output side by side in comparison but all the commands are formatted to be copied. The describe command outputs are also in code block rather than the screenshot. Please let me know what you think.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+
+### Delivery State
+
+![Delivery state](/assets/images/posts/2025-06-24-kafka-queue-02.png)
+
+Delivery count for each record is tracked and gets incremented every time a consumer acquires the record. There is a limit on the number of times a record can be retried to avoid getting stuck trying to deliver an unprocessable record indefinitely. This limit also can be reconfigured with broker configuration, <b>group.share.delivery.count.limit</b> with default value of 5.
+
+The delivery count determines which state the record should transition into later on. If the consumer releases the record or the lock duration has lapsed, it will go back into <b>Available</b> state as long as the delivery count has not reached the limit. 
+The record will go into <b>Acknowledged</b> state if it has been successfully processed by the consumer. If the consumer rejects the record or its delivery count has reached the limit, the record will go into <b>Archived</b> state. The <b>Archived</b> state represents a record that is removed from the queue because it was processed successfully or was unprocessable.
```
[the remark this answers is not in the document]
tinaselenge: You are right, it looks too confusing. I've changed and moved the last sentence to where I think it makes more sense.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+| 0    | Available    |
+| 2    | Acknowledged |
+| 4    | Archived     |
+
+Let’s take a look at the following example of a partition that a share group is subscribed to, which is called <b>Share Partition</b>:
+
+![Delivery state](/assets/images/posts/2025-06-24-kafka-queue-03.png)
+
+In this example, offset 2 is the start offset of the share group consuming from the partition. Records at offset 2 and 4 are currently <b>Acquired</b> (Acqui) for the first time, therefore delivery count is incremented to 1. However, the record at offset 3 is in <b>Available</b> (Avail) state with delivery count of 2, which means a consumer of the share group has attempted to deliver this record twice so far and it will be retried until the maximum delivery count is reached. The record at offset 5 has been processed successfully and <b>Acknowledged</b> (Ack) and the record at offset 6 is the next available record to be acquired by the share group, therefore it is the end offset for this share partition. This is an example of how these states would look like when stored in <code>__share_group_state</code>:
+```json
```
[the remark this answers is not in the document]
tinaselenge: I think I agree that it is too repetitive. I think the first example already shows different states and batches of states.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+            for (ConsumerRecord<String, String> record : records) {
+                try {
+                    if (record.key() != null && record.key().equals("skip")) {  // #1 
+                        System.out.println("Skipping the record - Offset: " + record.offset() +
+                                " Key: " + record.key() +
+                                " Value: " + record.value() +
+                                " Delivery count: " + record.deliveryCount().get());
+
+                        consumer.acknowledge(record, AcknowledgeType.RELEASE);
+                        break; // #2
```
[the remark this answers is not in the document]
tinaselenge: I put break to demonstrate the retry of this record faster and to show what happens if some records are not acknowledged at all in this case. I tried to explain why we put a break here.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,590 @@
...
+```java
+public class ShareGroupDemoImplicitAck {
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    @SuppressWarnings("InfiniteLoopStatement")
+    public static void main(final String[] args) {
+        Properties producerProps = new Properties();
+        producerProps.setProperty("bootstrap.servers", "localhost:9092");
+        producerProps.setProperty("key.serializer", "org.apache.kafka.common.serialization.StringSerializer");
+        producerProps.setProperty("value.serializer", "org.apache.kafka.common.serialization.StringSerializer");
```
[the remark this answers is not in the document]
tinaselenge: good catch!

tinaselenge:
Thank you so much to everyone who reviewed! 

> Do we want to talk about the metrics in the blog post?
https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka-Metrics

https://kafka.apache.org/documentation/#group_coordinator_monitoring

@showuon Good suggestion! However, I did not include metrics because they seem to be standard group coordinator metrics for both type of group coordinators, not many new metrics specifically for share group. And some metrics mentioned in the KIP seem to be not added yet. Since we don't have experience in using a share group, it was hard for me to tell which metrics would be most useful to monitor as some metrics seemed quite low level. Maybe this is something we can add once the feature is in production and is used more in the future.

mimaison:
In the [4.1.0 preview notes for Queues](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=364349960#QueuesforKafka(KIP932)PreviewReleaseNotes-Limitations), there are a few limitations noted, especially around compatibility with 4.0.0.

Since 4.1 is about to release it's probably worth testing and documenting the steps to disable queues on 4.0.0 after having followed your blog and verify you can upgrade to 4.1.

_posts/2025-06-24-kafka-queue.md
mimaison: Then it's probably something worth highlighting

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,585 @@
...
+    metadataVersion: 4.0-IV3
+  # ...
+```
+Since I'm running a single node cluster, `share.coordinator.state.topic.replication.factor` also had to be set to 1. We will cover more about this topic later.
+
+Using the Kafka command line tools, I will produce some records on the topic and consume those records using both consumer and share groups and show how they behave differently. 
+
+In this example, I will start with 3 share consumers, all joining the same share group called, `share-group` and then 3 regular consumers, also all joining the same consumer group called, "consumer-group". I will run the following command in different terminal windows with different pod names, e.g.  kafka-share-consumer-0,  kafka-share-consumer-1, kafka-share-consumer-2:
+```sh
+kubectl -n kafka run kafka-share-consumer-0 -ti --image=quay.io/strimzi/kafka:0.46.0-kafka-4.0.0 --rm=true --restart=Never -- bin/kafka-console-share-consumer.sh --bootstrap-server my-cluster-kafka-bootstrap:9092 --topic kafka-queue-topic --group share-group --property print.offset=true --property print.partition=true
```
mimaison, line 64: For these long commands, could we break them over multiple lines to improve readability? For example:
```
kubectl -n kafka run kafka-share-consumer-0 -ti --rm=true --restart=Never \
  --image=quay.io/strimzi/kafka:0.46.0-kafka-4.0.0 \
  -- bin/kafka-console-share-consumer.sh \
    --bootstrap-server my-cluster-kafka-bootstrap:9092 \
    --topic kafka-queue-topic --group share-group \
    --property print.offset=true \
    --property print.partition=true
```

tinaselenge:
@katheris @fvaleri @showuon Can you please take another look when you have time as I have made quite few changes since you approved? 

@scholzj @mimaison please let me know if you have few any further comments.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,703 @@
+---
+layout: post
+title:  "Using Queues for Kafka with Strimzi"
```
scholzj, line 3: I think it would be nice to use the full name in the title and in the first sentences ... I know the KIP is named _Queues for Kafka_ ... but still.

```suggestion
title:  "Using Queues for Apache Kafka with Strimzi"
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,703 @@
...
+date: 2025-06-24
+author: tina_selenge
+---
+
+The Queues for Kafka feature was introduced by [KIP-932](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka) and is in early access in Apache Kafka version 4.0 which is supported by Strimzi 0.46. In this blog, I’m going to introduce this feature and show how you can try it out with your Strimzi managed cluster.
```
scholzj, line 8:
```suggestion
The Queues for Kafka feature was introduced by [KIP-932](https://cwiki.apache.org/confluence/display/KAFKA/KIP-932%3A+Queues+for+Kafka#KIP932:QueuesforKafka) and is in early access in Apache Kafka version 4.0 which is supported from Strimzi 0.46. In this blog, I’m going to introduce this feature and show how you can try it out with your Strimzi managed cluster. 
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,703 @@
...
+
+This feature is only supported with Kafka clusters running in KRaft mode, since Zookeeper was removed in the Apache Kafka 4.0 release. It is also based on the new consumer rebalance protocol introduced by [KIP-848](https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol) that enhances stability, scalability and performance. 
+
+Traditional message queues and Kafka topics are significantly different in terms of design and use cases. A queue allows a group of consumers to read and process records in parallel to distribute the workload. It does not typically provide a strong ordering guarantee as records in a queue are consumed and processed independently. However it does provide exactly-once guarantee as each record is only read once and is deleted from the queue. States of individual records are usually tracked and acknowledged after being processed successfully to avoid prematurely removing records off the queue or to handle unprocessable records. This also means that records in a queue are not replayable, therefore a queue can only be consumed by a single group of consumers.
```
scholzj, line 12: > However it does provide exactly-once guarantee as each record is only read once and is deleted from the queue.

I would not say that traditional queueing system needs to have exactly once guarantees. At least once guarantees are pretty common. Given at the end of this section you say that Kafka does not support exactly-once for queues, maybe you can be less strict about it here? 😉

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,703 @@
...
+  name: kafka-queue-topic
+  labels:
+    strimzi.io/cluster: "my-cluster"
+spec:
+  partitions: 2
+  replicas: 1
+EOF
+```
+
+Since this feature is still in early access, as of Apache Kafka 4.0.0, you need to explicitly enable it by setting the following:
```
scholzj, line 39: Technically, there is nothing wrong with the order as you have it. But somehow I personally would first talk about preparing the Kafka cluster and only then create the topic ... just saying, I know it might be individual.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,703 @@
...
+    metadataVersion: 4.0-IV3
+  # ...
+```
+Since I'm running a single node cluster, `share.coordinator.state.topic.replication.factor` also had to be set to 1. We will cover more about this topic later.
+
+Using the Kafka command line tools, I will produce some records on the topic and consume those records using both consumer and share groups and show how they behave differently. 
+
+In this example, I will start with 3 share consumers, all joining the same share group called, `share-group` and then 3 regular consumers, also all joining the same consumer group called, "consumer-group". I will run the following command in different terminal windows with different pod names, e.g.  kafka-share-consumer-0,  kafka-share-consumer-1, kafka-share-consumer-2:
+```sh
+$ kubectl -n kafka run kafka-share-consumer-0 -ti --image=quay.io/strimzi/kafka:0.46.0-kafka-4.0.0 --rm=true --restart=Never \
```
scholzj, line 64: Maybe we can bump the image version here and in all the other cases to 0.47.0 to use the latest and greatest?

(You actually use 0.47.0 in some of the commands later which makes it even ore confusing)

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,703 @@
...
+| max.poll.records                       | Consumer | 500 | The maximum number of records returned in a single `poll()`. | 
+| share.acknowledgement.mode             | Consumer | implicit | Controls the acknowledgement mode for a share consumer. If set to implicit, must not use `acknowledge()`. Instead, delivery is acknowledged in the next `poll()` call. If set to explicit, must use  `acknowledge()` to acknowledge delivery of records. 
+
+The broker configurations can be set in your Kafka CR like shown in the example above. And the consumer configurations can be set in your client application, like shown in the Java API example.
+
+### The current limitations of the feature
+
+This feature is currently not recommended for production clusters as of Apache Kafka 4.0.0 but can be tried out with a test cluster as demonstrated above. If you are trying out this feature on your existing development/test cluster, note that you will not be able to upgrade to Apache Kafka 4.1 or later. You would have to disable this feature first and delete the state topic, in order to upgrade to a later version.
```
scholzj, line 699: > note that you will not be able to upgrade to Apache Kafka 4.1 or later

@ppatierno Curious how does this show up in your 4.1.0-RC PR. Did you test it? 😯

scholzj (approved):
Nice blog post, thanks. I left some comments, but feel free to treat them as _for consideration_ only. Apart from the version in the Strimzi images, I'm fine to have it published as is.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+```
+Subscribed partitions’ states are stored in batches as records get consumed and acknowledged in batches too. In the example above, you can see the batch of states, one of which is for offsets between 561 and 564. These offsets have the same delivery state and count therefore share the same state. In comparison, the offset 560 and 565 have different delivery states therefore listed in individual states.
+
+<i>Map of the states and enums:</i>
+
+| Enum | State |
+| :--: |:-------------|
+| 0    | Available    |
+| 2    | Acknowledged |
+| 4    | Archived     |
```
ppatierno, line 266: There are no 1 and 3? Acquired? or what else is missing?

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+                        consumer.acknowledge(record, AcknowledgeType.REJECT);  // #5
+                    } catch (Exception e ) {
+                        System.out.println("Failed to process the record- Offset: " + record.offset() + " Value: " + record.value());
+                        consumer.acknowledge(record, AcknowledgeType.RELEASE); // #6
+                    }
+                }
+                
+                consumer.acknowledge(record, AcknowledgeType.ACCEPT); // #7
+            }
+            consumer.commitSync(); // #7
```
ppatierno, line 378:
```suggestion
            consumer.commitSync(); // #8
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+  "recordNumber" : 16,
+  "recordValue" : "This is test message"
+}
+```
+
+The record at offset 361 was rejected because it hit an error mapping the record value into JSON therefore is considered a poison message. 
+The records at offset 362, 363, 364 and 365 were successfully processed and therefore accepted.
+The record at offset 366 was however released without getting processed therefore it was retried 5 times before getting archived and no longer being retried. The consumer then continued to the next record at offset 367.
+
+Let’s inspect the state of these records from the internal topic, __share_group_state just to understand the different states the records transitioned through:
```
ppatierno, line 493:
```suggestion
Let’s inspect the state of these records from the internal topic, `__share_group_state` just to understand the different states the records transitioned through:
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+kind: Kafka
+metadata:
+  # ...
+spec:
+  kafka:
+    config:
+      # ...
+      group.coordinator.rebalance.protocols: classic,consumer,share
+      unstable.api.versions.enable: true
+      share.coordinator.state.topic.replication.factor: 1
```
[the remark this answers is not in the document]
ppatierno: Reading the blog until the end I know that these will be mentioned in a table. Tbh I still think that you should mention them here (not using the table but just explaining their meaning).

_posts/2025-06-24-kafka-queue.md
ppatierno: No I didn't test it. I didn't know about this feature "blocking" the upgrade. @tinaselenge is it somewhere in official Kafka doc?

ppatierno (commented):
Very well written, nice blog post! I left some comments.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+--bootstrap-server my-cluster-kafka-bootstrap:9092 \
+--describe \
+--group consumer-group
+
+GROUP           TOPIC             PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG             CONSUMER-ID            HOST            CLIENT-ID
+consumer-group  kafka-queue-topic 1          13097           13097           0               CowB2m5kR5WDz_jIIoEDQw /10.244.0.135   console-consumer
+consumer-group  kafka-queue-topic 0          7903            7903            0               fzNIbX2kTk-Lmxb5SUh-HA /10.244.0.136   console-consumer
+```
+
+There are 3 members in this group, but only 2 of them are assigned to a partition each exclusively as the topic has only 2 partitions:
```
PaulRMellor, line 175:
```suggestion
This group had three members, but only two were assigned partitions because the topic had only two partitions.
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+    }
+
+    private static void handlePoisonRecord(ConsumerRecord<String, String> record, KafkaProducer<String, String> producer) {
+        producer.send(new ProducerRecord<>("dead-letter-queue-topic", record.key(), record.value()));
+    }
+}
+
+```
+
+#1. In this example, I have used the new consumer configuration added for this feature, `share.acknowledgement.mode` and set it to `explicit`. This configuration is set to `implicit` by default, which is why I didn’t need to set this configuration for the previous example.
```
PaulRMellor, line 394:
```suggestion
#1. In this example, I used the new consumer configuration added for this feature, `share.acknowledgement.mode`, and set it to `explicit`. This configuration is set to `implicit` by default, which is why I didn’t need to set this configuration for the previous example. 
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+    private static void handlePoisonRecord(ConsumerRecord<String, String> record, KafkaProducer<String, String> producer) {
+        producer.send(new ProducerRecord<>("dead-letter-queue-topic", record.key(), record.value()));
+    }
+}
+
+```
+
+#1. In this example, I have used the new consumer configuration added for this feature, `share.acknowledgement.mode` and set it to `explicit`. This configuration is set to `implicit` by default, which is why I didn’t need to set this configuration for the previous example. 
+
+#2. The consumer immediately releases a record if its key is set to `skip`, just to demonstrate the retry of delivery attempts. The record is fetched again in the next call to `poll()` and eventually hits the maximum delivery attempts. Then it is transitioned into <b>Archived</b> state because the client never processes it. Once archived, this record is no longer available for another delivery.
```
PaulRMellor, line 396:
```suggestion
#2. The consumer immediately releases a record if its key is set to `skip`, just to demonstrate the retry of delivery attempts. The record is fetched again in the next call to `poll()` and eventually hits the maximum delivery attempts. Then it is transitioned into an <b>Archived</b> state because the client never processes it. Once archived, this record is no longer available for another delivery. 
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+    }
+}
+
+```
+
+#1. In this example, I have used the new consumer configuration added for this feature, `share.acknowledgement.mode` and set it to `explicit`. This configuration is set to `implicit` by default, which is why I didn’t need to set this configuration for the previous example. 
+
+#2. The consumer immediately releases a record if its key is set to `skip`, just to demonstrate the retry of delivery attempts. The record is fetched again in the next call to `poll()` and eventually hits the maximum delivery attempts. Then it is transitioned into <b>Archived</b> state because the client never processes it. Once archived, this record is no longer available for another delivery. 
+
+#3. If the record's key is `skip`, it exits the loop so that we demonstrate the delivery retry and what happens if some records are not acknowledged during the poll.
```
PaulRMellor, line 398:
```suggestion
#3. If the record’s key is `skip`, the consumer exits the loop. This demonstrates how delivery retries work and what happens when some records remain unacknowledged during a `poll()` cycle.
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+
+#2. The consumer immediately releases a record if its key is set to `skip`, just to demonstrate the retry of delivery attempts. The record is fetched again in the next call to `poll()` and eventually hits the maximum delivery attempts. Then it is transitioned into <b>Archived</b> state because the client never processes it. Once archived, this record is no longer available for another delivery. 
+
+#3. If the record's key is `skip`, it exits the loop so that we demonstrate the delivery retry and what happens if some records are not acknowledged during the poll. 
+
+#4. Also when a record value cannot be mapped to a valid JSON object, the processing method throws a `JsonProcessingException` and the client handles it as a poison record. Since the dead letter queue is not supported yet, users have to implement this themselves. In this example, we send the records that cannot be processed to another topic `dead-letter-queue-topic`.
```
PaulRMellor, line 400:
```suggestion
#4. When a record value cannot be mapped to a valid JSON object, the processing method throws a `JsonProcessingException` and the client handles it as a poison record. Since the dead letter queue is not supported by this feature yet, users have to implement this themselves. In this example, we send the records that cannot be processed to another topic `dead-letter-queue-topic`. 
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+
+#5. Once a poison record is sent to the dead letter queue topic, the consumer client rejects it so that it gets archived immediately. 
+
+#6. If the client encounters an error not caused by the JSON mapping, it releases the record for another attempt because it could be a transient failure.
+
+#7. When no exception occurred during the processing, the consumer client accepts the record. Once all the records in the batch are processed and acknowledged individually, these states are stored locally in the consumer. 
+
+#8. Then the client has to call `commitSync()` or `commitAsync()` to commit the state to the internal topic, `__share_group_state`. In #2, it exits the loop for processing each record, and commits the state by calling `commitSync()`. The records in the batch that were not processed or acknowledged yet, will be presented to the consumer client again as part of the same acquisition, therefore their delivery count will not be incremented. Also note that share groups also do not support auto-commit.
```
PaulRMellor, line 408:
```suggestion
#8. The client calls `commitSync()` or `commitAsync()` to commit the state to the internal topic, `__share_group_state`. In #2, it exits the loop for processing each record, and commits the state by calling `commitSync()`. The records in the batch that were not processed or acknowledged yet, are presented to the consumer client again as part of the same acquisition and their delivery count is not incremented. Share groups do not support auto-commit.
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+
+Here is an example of a client application using [KafkaProducer](https://kafka.apache.org/40/javadoc/org/apache/kafka/clients/producer/KafkaProducer) to send records in valid and invalid format to `kafka-queue-topic` used for this demonstration:
```
PaulRMellor, line 410:
```suggestion
Here is an example of a client application using [KafkaProducer](https://kafka.apache.org/40/javadoc/org/apache/kafka/clients/producer/KafkaProducer) to send records in valid and invalid formats to the `kafka-queue-topic` used for this demonstration:
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+Skipping the record - Offset: 366 Key: skip Value: Record to skip processing Delivery count: 4
+Skipping the record - Offset: 366 Key: skip Value: Record to skip processing Delivery count: 5
+Processed record with offset 367: {
+  "isJson" : "true",
+  "recordNumber" : 16,
+  "recordValue" : "This is test message"
+}
+```
+
+The record at offset 361 was rejected because it hit an error mapping the record value into JSON therefore is considered a poison message.
```
PaulRMellor, line 489:
```suggestion
The record at offset 361 was rejected because it hit an error mapping the record value into JSON and was considered a poison message. 
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+Skipping the record - Offset: 366 Key: skip Value: Record to skip processing Delivery count: 5
+Processed record with offset 367: {
+  "isJson" : "true",
+  "recordNumber" : 16,
+  "recordValue" : "This is test message"
+}
+```
+
+The record at offset 361 was rejected because it hit an error mapping the record value into JSON therefore is considered a poison message. 
+The records at offset 362, 363, 364 and 365 were successfully processed and therefore accepted.
```
PaulRMellor, line 490:
```suggestion
The records at offset 362, 363, 364 and 365 were successfully processed and accepted.
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+The record at offset 366 was however released without getting processed therefore it was retried 5 times before getting archived and no longer being retried. The consumer then continued to the next record at offset 367.
```
PaulRMellor, line 491:
```suggestion
The record at offset 366 was released without being processed. It was retried five times before being archived and was not retried again. The consumer then continued to the next record at offset 367.
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+          "deliveryState": 4,
+          "deliveryCount": 1
+        }
+      ]
+    }
+  }
+}
+```
+
+The offset 361 that had a bad record was `Archived` (enum 4) state after the first delivery attempt, therefore was not retried again.
```
PaulRMellor, line 531:
```suggestion
The record at offset 361 was placed in the `Archived` (enum 4) state after the first delivery attempt, so it was not retried.
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+          "deliveryState": 2,
+          "deliveryCount": 1
+        }
+      ]
+    }
+  }
+}
+```
+
+The records at offset 364 and 365 were in <b>Acknowledged</b> (enum 2) after the first delivery attempt, as they were processed and acknowledged successfully by the consumer. (The offset 362 and 363 had the same state, just not included in the output)
```
PaulRMellor, line 588:
```suggestion
The records at offsets 364 and 365 were in the `Acknowledged` (enum 2) state after the first delivery attempt, as they had been processed and acknowledged successfully. (Offsets 362 and 363 were in the same state but are not shown in the output.)
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+          "deliveryState": 0,
+          "deliveryCount": 1
+        }
+      ]
+    }
+  }
+}
+```
+
+The record at offset 366 was in <b>Available</b> (enum 0) after the first delivery attempt. This is the record the consumer skipped from processing and simply released. However in the next screenshot, we see that delivery attempts have increased.
```
PaulRMellor, line 619:
```suggestion
The record at offset 366 was in the <b>Available</b> (enum 0) state after the first delivery attempt. This is the record the consumer skipped from processing and simply released. However in the following output, we see that delivery attempts have increased.
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+          "deliveryState": 4,
+          "deliveryCount": 5
+        }
+      ]
+    }
+  }
+}
+```
+
+After 4 attempts of delivery, the record was still in <b>Available</b> (enum 0) state but after the 5th attempt, it went into <b>Archived</b> (enum 4) state, which is why it was not retried again.
```
PaulRMellor, line 676:
```suggestion
After 4 attempts of delivery, the record was still in the <b>Available</b> (enum 0) state, but after the 5th attempt, it went into the <b>Archived</b> (enum 4) state, which is why it was not retried again. 
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+
+After 4 attempts of delivery, the record was still in <b>Available</b> (enum 0) state but after the 5th attempt, it went into <b>Archived</b> (enum 4) state, which is why it was not retried again. 
+
+### Broker and share configurations
+
+<i>Summary of the configurations mentioned in this blog:</i>
+
+| Configuration | Type | Default | Description |
+| :--: |:-------------| :-------------| :-------------|
+| group.coordinator.rebalance.protocols  | Broker   | classic,consumer | It should be set to `classic,consumer,share` to enable share group. |
```
PaulRMellor, line 684:
```suggestion
| group.coordinator.rebalance.protocols  | Broker   | classic,consumer | It must be set to `classic,consumer,share` to enable the share group. |
```

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+| group.share.session.timeout.ms         | Broker   | 45000 | The timeout to detect client failures when using the share group protocol. |
+| fetch.max.bytes                        | Broker   | 57671680 | The maximum number of bytes returned for a fetch request. |
+| max.poll.interval.ms                   | Consumer | 300000 | The maximum delay between invocations of `poll()` when using consumer group management. |
+| max.poll.records                       | Consumer | 500 | The maximum number of records returned in a single `poll()`. | 
+| share.acknowledgement.mode             | Consumer | implicit | Controls the acknowledgement mode for a share consumer. If set to implicit, must not use `acknowledge()`. Instead, delivery is acknowledged in the next `poll()` call. If set to explicit, must use  `acknowledge()` to acknowledge delivery of records. 
+
+The broker configurations can be set in your Kafka CR like shown in the example above. And the consumer configurations can be set in your client application, like shown in the Java API example.
```
PaulRMellor, line 696:
```suggestion
The broker configurations can be set in your Kafka CR, as shown in the example above. And the consumer configurations can be set in your client application, as shown in the Java API example.
```

PaulRMellor (approved):
Lots of useful information and a good example to explain the feature.
I guess it needs to be updated (or maybe a note) considering the change to **preview** in 4.1.0?
Suggestions mainly around readability, consistency, grammar and so on

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+
+<i>Summary of the configurations mentioned in this blog:</i>
+
+| Configuration | Type | Default | Description |
+| :--: |:-------------| :-------------| :-------------|
+| group.coordinator.rebalance.protocols  | Broker   | classic,consumer | It should be set to `classic,consumer,share` to enable share group. |
+| unstable.api.versions.enable           | Broker   | false | It should be set to true, in order to use this feature until it is production ready .|
+| share.coordinator.state.topic.replication.factor | Broker | 3 | Replication factor for the share-group state topic. Setting this greater than the cluster size will result in failure to create the topic and share group consumers will not work. |
+| group.share.record.lock.duration.ms    | Broker   | 30s | The record acquisition lock duration in milliseconds for share groups. |
+| group.share.partition.max.record.locks | Broker   | 200 | Share group record lock limit per share-partition. |
```
fvaleri, line 688: This is correct for Kafka 4.0 (see the other related comment from me).

fvaleri (commented):
@tinaselenge thanks for the updates.

I left a few more comments, but it looks mostly good.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+-- bin/kafka-share-groups.sh \
+--bootstrap-server my-cluster-kafka-bootstrap:9092 \
+--describe \
+--group share-group \
+--members
+
+GROUP           CONSUMER-ID            HOST            CLIENT-ID              ASSIGNMENT
+share-group     bSURQb3NTfSDGXQmgRrEeQ /10.244.0.131   console-share-consumer kafka-queue-topic:0,kafka-queue-topic:1
+share-group     HY1kEgMFShaumg8xSEc9sw /10.244.0.133   console-share-consumer kafka-queue-topic:0,kafka-queue-topic:1
+share-group     cvg5pnSlQ-qzlxocNkTplg /10.244.0.132   console-share-consumer kafka-queue-topic:0,kafka-queue-topic:1
```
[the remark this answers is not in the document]
tinaselenge: Which is why I think this output shows both partitions being assigned to each consumer member. There is no static assignment as in consumer 0 being assigned to one particular partition. I added a sentence about it.

_posts/2025-06-24-kafka-queue.md
```diff
@@ -0,0 +1,704 @@
...
+
+This feature is only supported with Kafka clusters running in KRaft mode, since Zookeeper was removed in the Apache Kafka 4.0 release. It is also based on the new consumer rebalance protocol introduced by [KIP-848](https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol) that enhances stability, scalability and performance. 
+
+Traditional message queues and Kafka topics are significantly different in terms of design and use cases. A queue allows a group of consumers to read and process records in parallel to distribute the workload. It does not provide a strong ordering guarantee as records in a queue are consumed and processed independently. However it typically provides exactly-once guarantee as each record is only read once and is deleted from the queue. States of individual records are usually tracked and acknowledged after being processed successfully to avoid prematurely removing records off the queue or to handle unprocessable records. This also means that records in a queue are not replayable, therefore a queue can only be consumed by a single group of consumers.
```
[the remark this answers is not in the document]
tinaselenge: I generalised it not to imply exactly-once is the default.

ppatierno (commented):
@tinaselenge I have just one last nit

PR #537, opened 2025-10-17
docs(glossary): reset internal links

PaulRMellor:
**web site**

Documentation page
Reset of glossary links

* [ ] Typo/minor fix
* [ ] Other


PaulRMellor:
I initially expected to revisit the source to reset the links, but this script achieves what we need: keeping the glossary self-contained and version-agnostic, while linking to the latest docs in the same format used elsewhere in the site.

If there’s a simpler or cleaner way to handle this, I’m happy to adjust.

PR #577, opened 2026-05-04
Fix incorrect property paths in KafkaMirrorMaker2 v1 API migration docs

### LINKED ISSUE #576: The KafkaMirrorMaker2 v1 API migration documentation (section 30.7.4) contains incorrect property paths.
## Description
The KafkaMirrorMaker2 v1 API migration documentation ([section 30.7.4](https://strimzi.io/docs/operators/latest/deploying#ref-kafkamirrormaker2-v1-changes-str)) contains incorrect property paths.

### Error #1: Wrong Target Property Paths
Documentation incorrectly states:
- `.spec.groupId`
- `.spec.configStorageTopic`
- `.spec.offsetStorageTopic`
- `.spec.statusStorageTopic`

Should be:
- `.spec.target.groupId`
- `.spec.target.configStorageTopic`
- `.spec.target.offsetStorageTopic`
- `.spec.target.statusStorageTopic`

### Error #2: Incorrect Reference to .spec.config
Documentation states: 
"set from `.spec.config.group.id `"
"set from `.spec.config.config.storage.topic`"
"set from `.spec.config.offset.storage.topic `"
"set from `.spec.config.status.storage.topic`"

Problem: `.spec.config` does not exist in KafkaMirrorMaker2 CRD

Should be: 
"set from `.spec.clusters[].config.group.id` of the target cluster"
"set from `.spec.clusters[].config.config.storage.topic` of the target cluster"
"set from `.spec.clusters[].config.offset.storage.topic` of the target cluster"
"set from `.spec.clusters[].config.status.storage.topic` of the target cluster"

## Affected Versions
All versions from 0.49.0 through latest

Meenu-Mariya:
Fixes #576

### Type of change

_Select the type of your PR_

* [x] Typo/minor fix
* [ ] Other

### Description
Fixed incorrect property paths in the KafkaMirrorMaker2 v1 API migration documentation (section 30.7.4).

Error #1: Wrong Target Property Paths
Documentation incorrectly stated properties at .spec level instead of .spec.target:

Changed .spec.groupId → .spec.target.groupId
Changed .spec.configStorageTopic → .spec.target.configStorageTopic
Changed .spec.offsetStorageTopic → .spec.target.offsetStorageTopic
Changed .spec.statusStorageTopic → .spec.target.statusStorageTopic

Error #2: Incorrect Reference to .spec.config
Documentation incorrectly referenced .spec.config which does not exist in KafkaMirrorMaker2 CRD.

Before:
"set from .spec.config.group.id"
"set from .spec.config.config.storage.topic"
"set from .spec.config.offset.storage.topic"
"set from .spec.config.status.storage.topic"

After:
"set from .spec.clusters[].config.group.id of the target cluster"
"set from .spec.clusters[].config.config.storage.topic of the target cluster"
"set from .spec.clusters[].config.offset.storage.topic of the target cluster"
"set from .spec.clusters[].config.status.storage.topic of the target cluster"

### DIFF
--- a/docs/operators/0.49.0/deploying-book.html
+++ b/docs/operators/0.49.0/deploying-book.html
@@ -35343,16 +35343,16 @@ If <code>pause: true</code> was set, it is converted to <code>state: paused</cod
 <div class="ulist">
 <ul>
 <li>
-<p><code>.spec.groupId</code>: set from <code>.spec.config.group.id</code>, or default to <code>connect-cluster</code>.</p>
+<p><code>.spec.target.groupId</code>: set from <code>.spec.clusters[].config.group.id</code> of the target cluster, or default to <code>connect-cluster</code>.</p>
 </li>
 <li>
-<p><code>.spec.configStorageTopic</code>: set from <code>.spec.config.config.storage.topic</code>, or default to <code>connect-cluster-configs</code>.</p>
+<p><code>.spec.target.configStorageTopic</code>: set from <code>.spec.clusters[].config.config.storage.topic</code> of the target cluster, or default to <code>connect-cluster-configs</code>.</p>
 </li>
 <li>
-<p><code>.spec.offsetStorageTopic</code>: set from <code>.spec.config.offset.storage.topic</code>, or default to <code>connect-cluster-offsets</code>.</p>
+<p><code>.spec.target.offsetStorageTopic</code>: set from <code>.spec.clusters[].config.offset.storage.topic</code> of the target cluster, or default to <code>connect-cluster-offsets</code>.</p>
 </li>
 <li>
-<p><code>.spec.statusStorageTopic</code>: set from <code>.spec.config.status.storage.topic</code>, or default to <code>connect-cluster-status</code>.</p>
+<p><code>.spec.target.statusStorageTopic</code>: set from <code>.spec.clusters[].config.status.storage.topic</code> of the target cluster, or default to <code>connect-cluster-status</code>.</p>
 </li>
 </ul>
 </div>
--- a/docs/operators/0.49.0/full/deploying.html
+++ b/docs/operators/0.49.0/full/deploying.html
@@ -35785,16 +35785,16 @@ If <code>pause: true</code> was set, it is converted to <code>state: paused</cod
 <div class="ulist">
 <ul>
 <li>
-<p><code>.spec.groupId</code>: set from <code>.spec.config.group.id</code>, or default to <code>connect-cluster</code>.</p>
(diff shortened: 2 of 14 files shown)


scholzj:
Thanks for the Pr. however, documentation has to be fixed in https://github.com/strimzi/strimzi-kafka-operator/tree/main/documentation and not in this repository.

scholzj:
Also, what you desribe as the second issue seems like an issue, but the fix does not seem correct. So you should probably start by sharing a link to where the wrong documentations is in the first place so that we can decide if it is wrong and how to fix it.

Meenu-Mariya:
> Also, what you desribe as the second issue seems like an issue, but the fix does not seem correct. So you should probably start by sharing a link to where the wrong documentations is in the first place so that we can decide if it is wrong and how to fix it.

This is the wrong documentation path:
https://strimzi.io/docs/operators/latest/deploying#ref-kafkamirrormaker2-v1-changes-str. I have referenced it in the issue also #576

scholzj:
You are right, that looks like a bad copy-paste from Kafka Connect. However, this part was already removed from the documentation in the main branch. So there is nothing to fix there. I do not know if we want to fix it here as that would create inconsistencies with the actual release.

Meenu-Mariya:
> You are right, that looks like a bad copy-paste from Kafka Connect. However, this part was already removed from the documentation in the main branch. So there is nothing to fix there. I do not know if we want to fix it here as that would create inconsistencies with the actual release.

Thank you for the clarification. I understand this part was removed from main.

But since this error exists in the published documentation on strimzi.io (which users still reference), would you prefer:

- Fix in [strimzi.github.io](https://github.com/strimzi/strimzi.github.io/) (the published docs) 

- Fix in release branches: from release-0.49.x to release-1.0.x in [strimzi-kafka-operator](https://github.com/strimzi/strimzi-kafka-operator/)

- No fix - Leave it as-is.

scholzj:
I think that needs to be decided in the triage. Fixing it in the branches might not be very helpful. There might be 1.0.1 one day. But there definitely won't be patch releases for 0.5x for example. So no point in fixing it there. changing it only on the website diverges things from the releases. Which is not great either. But the issue sucks, so maybe it is the best we can do.

scholzj (approved):

see-quick (approved):

ppatierno (approved):

Frawless (approved):

scholzj:
We discussed this offline and decided to merge this. In case we do some 1.0.1 patch release, feel free to also open a Pr against the `release-1.0.x` branch in the operators repo (https://github.com/strimzi/strimzi-kafka-operator/tree/release-1.0.x).

PR #579, opened 2026-05-06
Add Strimzi Access Operator 0.3.0 after release

im-konge:
### Type of change

* New release

### Description

This PR adds Strimzi Access Operator 0.3.0 to the website after the release.


scholzj (commented):
I think this is published only as OCI artifact? (The Operator and Drain clenaer are published here for historical reasons.)

im-konge:
Going to close this as we should not add Access Operator to the index.yaml.

